Issue #12, opened 2018-10-08
Foresee for shared-access-key authentication through attributes


SamVanhoutte:
**Problem description**
Quite often (especially in backend integration), a secret key is needed for simple and easy to configure API authentication.  This feature should help in getting a fast solution.

**Solution description**
A declarative way to tag operations on API controllers with an Authorize attribute that refers to a SharedAccessPolicy is required will allow users to specify the needed authentication handler to authorize calls to the operation.

A sample implementation is already made by @SamVanhoutte (inspired by Frederik Gheysels)

tomkerkhove:
@SamVanhoutte Can you elaborate on the sample implementation that you've used?

This approach should be using the `ISecretProvider` (https://github.com/arcus-azure/arcus.security/issues/20) in my opinion.

tomkerkhove:
The sample Sam mentioned is available here - https://gist.github.com/tomkerkhove/e44e4e84f1f85acc37540fa941a5401b

tomkerkhove:
We'll introduce a new attribute that inherits from `ActionFilterAttribute` which will handle this authentication.

It should provide the capability to :
- Configure name of the header to check
- Configure name of the secret that contains the key

The attribute will use CachedSecretProvider and if the key is not correct, it will return an `HTTP 401 Unauthorized`.

We want to be able to plug this in the general pipeline via startup and will see if we can use it on a per-controller and/or per-operation bases if we can inject dependencies. _([Good read](https://stackoverflow.com/questions/36109052/inject-service-into-action-filter/36109690))_

Thanks for the feedback & input @stijnmoreels &	@fgheysels!

tomkerkhove:
Any suggestions for a clear name? How about `SharedAccessKeyAuthenticationAttribute`?

tomkerkhove:
Implemented and released in https://github.com/arcus-azure/arcus.webapi/releases/tag/v0.1.0-preview-2.

@stijnmoreels would you mind providing some documentation on how to use this feature?

PR #32, opened 2019-03-13
FEAT: add attribute for configurable shared access key authentication

stijnmoreels:
Adds an attribute to configure easily authentication with a shared key.
It provides a configurable request header and secret name.

 #12


stijnmoreels:
Added some unit tests but not sure how I should test the attribute on a controller. It almost looks like a AOP test. Any ideas?

tomkerkhove:
You can start it via a TestServer and call it like that in a unit test, if you want.

Example can be found here - https://github.com/arcus-azure/arcus.eventgrid/blob/master/src/Arcus.EventGrid.Tests.Unit/Security/WebhookValidationTests.cs

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -0,0 +1,104 @@
...
+        private readonly string _headerName, _secretName;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SharedAccessKeyAuthenticationAttribute"/> class.
+        /// </summary>
+        /// <param name="headerName">The name of the request header which value must match the stored secret.</param>
+        /// <param name="secretName">The name of the secret that's being retrieved using the <see cref="ISecretProvider.Get"/> call.</param>
+        public SharedAccessKeyAuthenticationAttribute(string headerName, string secretName)
+        {
+            if (String.IsNullOrWhiteSpace(headerName))
```
tomkerkhove: Use guards instead?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            _headerName = headerName;
+            _secretName = secretName;
+        }
+        
+        /// <inheritdoc />
+        public override async Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next)
+        {
+            if (context == null)
```
tomkerkhove: Use guards instead?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                throw new ArgumentException(
+                    "Invalid action context given without any HTTP request services", 
+                    nameof(context.HttpContext.RequestServices));
+            }
+
+            var userDefinedSecretProvider = context.HttpContext.RequestServices.GetService<ISecretProvider>();
+            if (userDefinedSecretProvider == null)
+            {
+                throw new KeyNotFoundException(
```
tomkerkhove: Is KeyNotFoundException actually correct here?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -0,0 +1,104 @@
...
+                    $"No configured {nameof(ISecretProvider)} implementation found in the request service container. "
+                    + "Please configure such an implementation (ex. in the Startup) of your application");
+            }
+
+            var cachedSecretProvider = new CachedSecretProvider(userDefinedSecretProvider);
```
tomkerkhove: We are creating a new cached provider every time so it's not really cached, is it? Can't we just get an instance from ICachedSecretProvider instead?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -0,0 +1,104 @@
...
+            string foundSecret = await cachedSecretProvider.Get(_secretName);
+            if (foundSecret == null)
+            {
+                throw new AuthenticationException(
```
tomkerkhove: Do we have a SecretNotFoundException here or in Arcus Security?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -0,0 +1,104 @@
...
+                    $"No secret found with name {_secretName} in {nameof(ISecretProvider)} implementation {userDefinedSecretProvider.GetType().Name}");
+            }
+
+            bool noRequestSecretHeaderFound =
```
tomkerkhove: Best to do this check before we get the secret, maybe we don't even need the secret as the header is not there?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -0,0 +1,104 @@
...
+                throw new AuthenticationException(
+                    $"No secret found with name {_secretName} in {nameof(ISecretProvider)} implementation {userDefinedSecretProvider.GetType().Name}");
+            }
+
+            bool noRequestSecretHeaderFound = 
+                context.HttpContext.Request.Headers
+                       .TryGetValue(_headerName, out StringValues requestSecretHeaders) == false;
+
+            bool anyRequestSecretHeaderDoesntMatch = 
+                requestSecretHeaders.Any(h => !String.Equals(h, foundSecret));
```
tomkerkhove: what does `h` stand for? Pref to not have abbreviations

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: Ah, yes. I've forgot that this was the code convention here.

stijnmoreels: Ah, yes. I've forgot that this was the code convention here.

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: Ah, ok. Didn't thought that this was part of the code style guide.
Usually with lambda expressions on sequences, we use the first letter of the type in the sequence.

So in this case it's a series of request headers, therefore the "H". But I'll rename it if you want.

tomkerkhove: Personally I'm not a fan of this because it's not really clear what it represents but I'll leave it up to you!

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: We could but then we limit the users to only use cached providers. Is that something we should force on users?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: Yes, we have! I'll use that.

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: Is used for missing values for a given key. So in this case I think it's correctly used.

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
tomkerkhove: Personally I would do that to stimulate good consumer behavior - What do you think @SamVanhoutte 	?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
tomkerkhove: 👍

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
tomkerkhove: Great, thanks! :+1:

src/Arcus.WebApi.Security/Arcus.WebApi.Security.csproj
```diff
@@ -0,0 +1,23 @@
...
+      <Company>Arcus</Company>
+      <RepositoryType>Git</RepositoryType>
+      <PackageTags>Azure;WebAPI;App Services;Web App;Web;API</PackageTags>
+      <Description>Provides capabilities to easily build Web APIs running in Azure.</Description>
+      <Copyright>Copyright (c) Arcus</Copyright>
+      <PackageLicenseUrl>https://github.com/arcus-azure/arcus.webapi/blob/master/LICENSE</PackageLicenseUrl>
+      <PackageProjectUrl>https://github.com/arcus-azure/arcus.webapi</PackageProjectUrl>
+      <RepositoryUrl>https://github.com/arcus-azure/arcus.webapi</RepositoryUrl>
+      <PackageIconUrl>https://raw.githubusercontent.com/arcus-azure/arcus/master/media/arcus.png</PackageIconUrl>
+      <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
tomkerkhove: Can you include the docs like you PRed for our other repos please?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            Guard.NotNullOrWhitespace(headerName, nameof(headerName), "Header name cannot be blank");
+            Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Secret name cannot be blank");
+
+            _headerName = headerName;
+            _secretName = secretName;
+        }
+        
+        /// <inheritdoc />
+        public override async Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next)
```
tomkerkhove: This looks a lot better, thank you!

src/Arcus.WebApi.Security/Arcus.WebApi.Security.csproj
stijnmoreels: Ah, yes.

stijnmoreels:
@tomkerkhove I tried to write both tests and prod code as correct as possible with the style guidelines.

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        [InlineData(null, "not empty or whitespace")]
+        [InlineData("", "not empty or whitespace")]
+        [InlineData(" ", "not empty or whitespace")]
+        [InlineData("not empty or whitespace", null)]
+        [InlineData("not empty or whitespace", "")]
+        [InlineData("not empty or whitespace", " ")]
+        public void SharedAccessKeyAttributeAndFilter_WithNotPresentHeaderNameAndOrSecretName_ShouldFailWithArgumentException(
+            string headerName,
+            string secretName)
+        {
```
tomkerkhove: Shouldn't this be splitted into two tests actually?

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
+        /// </summary>
+        public void Dispose()
+        {
+            _testServer.Dispose();
+        }
+    }
+    
+    [ApiController]
+    public class SharedAccessKeyAuthenticationController : ControllerBase
```
tomkerkhove: Let's only use one class per file

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            // Arrange
+            string secretValue = $"secret-{Guid.NewGuid()}";
+            _testServer.AddService<ISecretProvider>(new InMemorySecretProvider((SecretName, secretValue)));
+
+            // Act
+            using (HttpResponseMessage response = await SendAuthorizedHttpRequestWithHeader(HeaderName, secretValue))
+            {
+                // Assert
+                Assert.NotEqual(HttpStatusCode.Unauthorized, response.StatusCode);
```
tomkerkhove: I've misread this one, can't we assert on Ok or something instead?

src/Arcus.WebApi.Security/Arcus.WebApi.Security.csproj
tomkerkhove: Thanks!

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
tomkerkhove: I think this is resolved, right @stijnmoreels ?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
tomkerkhove: I think we should stimulate them to use cached or not?

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
stijnmoreels: Could be, but as testing a single 'unit' I guess it would be logical to test the inputs together on default argument exceptions.

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
stijnmoreels: Well, about that. Would that make the test more readable? SInce then you have to look in different files for a single test while the fixture setup is now in the same collection of tests.

I know we often have different opinions on this (both in prod- als test-code)

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
stijnmoreels: Well, it returned something likee `ContentTypeNotSupported` as status code. But I didn't quite found any reason (even when I explictily allowed all content-types).

But then, I thought about the what I wanted to test, and I don't want to test what status code the fixture (we ourselves created during the test setup) returns. But rather what the `SharedAccessKeyAuthenticationFIlter` NOT returned.

Therefore I find testing on NOT `Unauthorized` more correct.

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
stijnmoreels: But if you want this I can change it. Seperate in a `.\Fixtures\` folder?

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: Ah, yes we could first test if that returned anything and go for the `ISecretProvider` if not.

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: Ah yes, is switched now to first look in the request headers.

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: That's done.

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
tomkerkhove: I'm not sure if I agree as you are testing validation for 2 classes in one test?

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
tomkerkhove: Fair enough!

src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
tomkerkhove: Looks perfect, thanks!

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
stijnmoreels: Ah, ow; didn't see that. Yes. I'll split that.

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
tomkerkhove: It's more to be consistent with all the other files/classes as they don't do that as well. Next to that it's also to make the class more discoverable.

I'll leave it up to you to decide but I'm not sure how it's directly related to the tests itself as well, but that's ok.

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
tomkerkhove: No problem, that's why we have reviews!

tomkerkhove:
Green build: https://codit.visualstudio.com/Arcus/_build/results?buildId=31946

Issue #14, opened 2018-10-08
Provide a global exception handler


tomkerkhove:
Provide a global exception handler that can be plugged in the middleware and catches unhandled exceptions. These are wrapped in an `application/problem+json` response.

Adding @massimoc given he will probably want to add more input.

tomkerkhove:
@MassimoC For unhandled exceptions I'm actually not sure if we should return an application/problem+json, would that be recommended? I presume it should be?

tomkerkhove:
We should determine what the state of .NET is and otherwise use https://github.com/khellang/Middleware as a basis.

MassimoC:
The current .NET implementation is not enough. The middleware of khellang is seems to be the best approach as today.

fgheysels:
I'm picking up this issue, and I also wonder if we should use a problem+json response when an unhandled exception is being caught.  I mean, you would want to expose as less implementation details on what went wrong to the outside, do you ?

tomkerkhove:
That's a good point and a I agree on that, we should just return an HTTP 500 without details towards the client.

tomkerkhove:
Implemented and released in https://github.com/arcus-azure/arcus.webapi/releases/tag/v0.1.0-preview-exception-middelware.

@fgheysels will provide docs later next week/this month.

PR #33, opened 2019-04-16
Add middleware to log unhandled exceptions to various targets

fgheysels:
## Description
This is the minimal functionality that I require for the moment in an ASP.NET API.   This functionality allows to log exceptions to multiple targets.
It surely needs to be extended to allow to add correlation-id's or other (custom) information.

## Documentation

I don't know what we're doing regarding documentation ?  
Anyway, usage of this middleware:

In `Startup.Configure` , this middleware can be added to the pipeline:

```csharp
public void Configure(IApplicationBuilder app, IHostingEnvironment env)
{
}
```
The above statement will make sure that all unhandled exceptions are logged to all configured loggers.
To get the exceptions logged in AppInsights, logging should be configured like this when configuring the `IWebHost` (in Program.cs):

```csharp
WebHost.CreateDefaultBuilder(args)
              .UseApplicationInsights()
              .ConfigureLogging(loggers => loggers.AddApplicationInsights())
              .UseStartup<Startup>();
```

To be able to use the `AddApplicationInsights` extension method, the `Microsoft.Extensions.Logging.ApplicationInsights` package must be installed.

If the parameter-less `AddApplicationInsights` method is used, the configurationsetting `ApplicationInsights:TelemetryKey` must be specified and the value of the telemetry-key of the ApplicationInsights resource must be set.

#14

### DIFF
--- a/src/Arcus.WebApi.Logging/Arcus.WebApi.Logging.csproj
+++ b/src/Arcus.WebApi.Logging/Arcus.WebApi.Logging.csproj
@@ -6,7 +6,7 @@
     <Company>Arcus</Company>
     <RepositoryType>Git</RepositoryType>
     <PackageTags>Azure;WebAPI;App Services;Web App;Web;API</PackageTags>
-    <Description>Provides capabilities to easily build Web APIs running in Azure.</Description>
+    <Description>Provides capabilities for easily integrating logging when building Web APIs running in Azure.</Description>
     <Copyright>Copyright (c) Arcus</Copyright>
     <PackageLicenseUrl>https://github.com/arcus-azure/arcus.webapi/blob/master/LICENSE</PackageLicenseUrl>
     <PackageProjectUrl>https://github.com/arcus-azure/arcus.webapi</PackageProjectUrl>
--- a/src/Arcus.WebApi.Logging/Arcus.WebApi.Logging.csproj
+++ b/src/Arcus.WebApi.Logging/Arcus.WebApi.Logging.csproj
@@ -1,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>    
+    <Authors>Arcus</Authors>
+    <Company>Arcus</Company>
+    <RepositoryType>Git</RepositoryType>
+    <PackageTags>Azure;WebAPI;App Services;Web App;Web;API</PackageTags>
+    <Description>Provides capabilities to easily build Web APIs running in Azure.</Description>
+    <Copyright>Copyright (c) Arcus</Copyright>
+    <PackageLicenseUrl>https://github.com/arcus-azure/arcus.webapi/blob/master/LICENSE</PackageLicenseUrl>
+    <PackageProjectUrl>https://github.com/arcus-azure/arcus.webapi</PackageProjectUrl>
+    <RepositoryUrl>https://github.com/arcus-azure/arcus.webapi</RepositoryUrl>
+    <PackageIconUrl>https://raw.githubusercontent.com/arcus-azure/arcus/master/media/arcus.png</PackageIconUrl>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.2.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
+  </ItemGroup>
+
+</Project>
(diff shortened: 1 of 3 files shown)


tomkerkhove:
THanks @fgheysels! In terms of docs I recommend adding them to `/docs` similar for other Arcus projects - https://github.com/arcus-azure/arcus.security/tree/master/docs

Would you mind moving the docs there please?

src/Arcus.WebApi.Logging/Arcus.WebApi.Logging.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>    
+    <Authors>Arcus</Authors>
+    <Company>Arcus</Company>
+    <RepositoryType>Git</RepositoryType>
+    <PackageTags>Azure;WebAPI;App Services;Web App;Web;API</PackageTags>
+    <Description>Provides capabilities to easily build Web APIs running in Azure.</Description>
```
tomkerkhove: Given this is the logging package we should provide a more suited description for it.

How about this?
```suggestion
    <Description>Provides capabilities for easily integrating logging when building Web APIs running in Azure.</Description>
```

src/Arcus.WebApi.Logging/Arcus.WebApi.Logging.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <Copyright>Copyright (c) Arcus</Copyright>
+    <PackageLicenseUrl>https://github.com/arcus-azure/arcus.webapi/blob/master/LICENSE</PackageLicenseUrl>
+    <PackageProjectUrl>https://github.com/arcus-azure/arcus.webapi</PackageProjectUrl>
+    <RepositoryUrl>https://github.com/arcus-azure/arcus.webapi</RepositoryUrl>
+    <PackageIconUrl>https://raw.githubusercontent.com/arcus-azure/arcus/master/media/arcus.png</PackageIconUrl>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
tomkerkhove: Can you add `GenerateDocumentationFile` as well, please? ([example](https://github.com/arcus-azure/arcus.security/blob/master/src/Arcus.Security.Secrets.AzureKeyVault/Arcus.Security.Secrets.AzureKeyVault.csproj#L16))

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
```diff
@@ -0,0 +1,38 @@
...
+            }
+            catch (Exception ex)
+            {
+                HandleException(context, ex, loggerFactory);
+            }
+        }
+
+        private static void HandleException(HttpContext context, Exception ex, ILoggerFactory loggerFactory)
+        {
+            var logger = loggerFactory.CreateLogger("");
```
tomkerkhove: This looks odd, is this intentional? Is the string to reflect a "section" of the logging? Maybe use "Infrastructure instead of configurable with empty string as default?

fgheysels: This is odd indeed :)
It seems that you need to provide a category for the logger or use a type parameter which will give you then a `ILogger<T>`.   Since this is a package that will be integrated in other solutions, I couldn't come up with a suitable category-name here, but I'll solve this in another way.

fgheysels: I've modified it so that consumers of this package can specify the category-name when they're adding the middleware to their pipeline.
Like this:

```csharp
Func<string> getCategoryName = () => "My Category";

app.UseMiddleware<Arcus.WebApi.Logging.ExceptionHandlingMiddleware>(args: new object[] { getCategoryName });
```
or this:

```csharp
app.UseMiddleware<Arcus.WebApi.Logging.ExceptionHandlingMiddleware>(args: new object[] {"My Category" });
```

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+        { }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExceptionHandlingMiddleware"/> class.
+        /// </summary>
+        /// <param name="next"></param>
+        /// <param name="getLoggingCategory">The function that returns the category-name that must be used by the logger
+        /// when writing log messages.</param>
+        public ExceptionHandlingMiddleware(RequestDelegate next, Func<string> getLoggingCategory)
+        {
```
tomkerkhove: Maybe good to guard for getLoggingCategory being null?

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
tomkerkhove: Looks good!

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
fgheysels: Well, I'd like to avoid throwing an exception for that so I've solved it during logging itself.  If categoryName resolves to null, it gets replaced with an empty string.

tomkerkhove: That's also ok for us

src/Arcus.WebApi.Logging/Arcus.WebApi.Logging.csproj
fgheysels: Done.

Issue #16, opened 2018-10-08
Provide capability to track unhandled exceptions in Application Insights


tomkerkhove:
Provide capability to track unhandled exceptions in Application Insights.

This depends on #14.

tomkerkhove:
This is actually already done via #14 which uses `ILogger` - Can you confirm this @fgheysels?

fgheysels:
> This is actually already done via #14 which uses `ILogger` - Can you confirm this @fgheysels?

Yes this is done by the functionality mentionned in #14.  One of my pull-requests mentionned #16 as well, so this one can be closed imo.

tomkerkhove:
Perfect, thanks @fgheysels !

PR #41, opened 2019-05-10
Use more reading-friendly headers in docs

### LINKED ISSUE #12: Foresee for shared-access-key authentication through attributes

tomkerkhove:
Fixes #12 
Relates to #16

### DIFF
--- a/docs/features/logging.md
+++ b/docs/features/logging.md
@@ -3,7 +3,7 @@ title: "Logging unhandled exceptions via ASP.NET Core middleware"
 layout: default
 ---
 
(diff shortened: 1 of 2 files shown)


stijnmoreels:
Ah OK, now I see

Issue #39, opened 2019-04-27
Provide certificate-based authentication through middleware


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Quite often we need to enforce tight security on an API where certificates are often used.

**Describe the solution you'd like**
A declarative way to configure client certificates that are allowed by using middleware which enforces this on all endpoints.

**Additional context**
This relates to #31 which will provide attribute-based decoration which will be very similar. Both of these should use the same shared core.

tomkerkhove:
Closed via #43

Issue #31, opened 2019-01-29
Provide certificate-based authentication through attributes


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Quite often we need to enforce tight security on an API where certificates are often used.

**Describe the solution you'd like**
A declarative way to tag operations on API controllers with an Authorize attribute that defines what client certificates that are allowed.

MassimoC:
An example of certificate validation can be found here: https://docs.microsoft.com/en-us/azure/app-service/app-service-web-configure-tls-mutual-auth

stijnmoreels:
Could be fun to play with. As test we should generate those certificates with different subject names, expiration dates, ...

@tomkerkhove is this somewhat the same approach as the **Shared Access Key** attributes?

tomkerkhove:
Yes, but it all depends on the way it should work and required configuration so I think we need more insights first.

fgheysels:
> An example of certificate validation can be found here: https://docs.microsoft.com/en-us/azure/app-service/app-service-web-configure-tls-mutual-auth

A bit disappointing that the ASP.NET sample is using some old ASP.NET WebForms code.  That's like, 15 years old or so. :)

fgheysels:
> **Is your feature request related to a problem? Please describe.**
> Quite often we need to enforce tight security on an API where certificates are often used.
> 
> **Describe the solution you'd like**
> A declarative way to tag operations on API controllers with an Authorize attribute that defines what client certificates that are allowed.

So, how do you see this in action / used ?  Something like this :

```csharp
[ClientCertificateValidation(CertificateThumbPrint="abcdefghi")]
public class MyController
{
}
```
?
I am doing something similar in another project (no client cert validation, but some additional/custom authorization), so I might be fiddling around with it a bit more to get something like the above working.

tomkerkhove:
There are a couple of ways to achieve this and I'm open for feedback!

Personally, I would use the same approach like the shared access key attribute where you can either use an attribute or plug it in the middleware.

fgheysels:
Indeed, I was also wondering if you want to have it as fine grained as this.  I think it would be useful if you could just plug it in as middleware so that the certificate validation is done on the entire api instead of having to apply it on each controller separately.

tomkerkhove:
After discussion we see value in the following approaches:
- **Middleware** to enforce this on all operations which should be used as a default approach
- **Attributes** in case you only want to secure a subset of endpoints with this approach
[...]

PR #43, opened 2019-06-12
FEAT: add certificate based authentication filter & attribute

stijnmoreels:
One requires a easy way to allow only a certain amount of client certificates on global and local level. This commit provides this way by using a global filter and a local attribute.

#31
#39


src/Arcus.WebApi.Unit/Hosting/TestApiServer.cs
```diff
@@ -71,5 +81,43 @@ protected override IWebHostBuilder CreateWebHostBuilder()
...
+        /// </summary>
+        /// <param name="clientCertificate">The client certificate.</param>
+        public void SetClientCertificate(X509Certificate2 clientCertificate)
+        {
+            Guard.NotNull(clientCertificate, nameof(clientCertificate));
+
+            _clientCertificate = clientCertificate;
+        }
+
+        private class CertificateConfiguration : IStartupFilter
```
tomkerkhove: Let's not use inner classes here

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationAttributeTests.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Arcus.WebApi.Unit.Hosting;
+using Arcus.WebApi.Unit.Security.Doubles;
+using Xunit;
+using static Arcus.WebApi.Unit.Security.Authentication.CertificateAuthenticationController;
+
+namespace Arcus.WebApi.Unit.Security.Authentication
```
tomkerkhove: Can we also add some tests where the certificate is enforced globally please?

src/Arcus.WebApi.Security/Authentication/X509Validation.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Arcus.WebApi.Security.Authentication 
+{
+    /// <summary>
+    /// Represents which value of the client <see cref="X509Certificate2"/> should be validated in the <see cref="CertificateAuthenticationFilter"/>.
+    /// </summary>
+    public enum X509Validation
```
tomkerkhove: This is interesting - This enum is not really validating the cert but more defining the requirements, no?

How about this:
```suggestion
    public enum X509ValidationRequirement
```

tomkerkhove:
I've had a quick look and looks very good! Before I drill deeper I'd like to see some test for the global enforcement and see how that feels from a consumer side.

stijnmoreels:
> I've had a quick look and looks very good! Before I drill deeper I'd like to see some test for the global enforcement and see how that feels from a consumer side.

Yes, of course.
Thanks!

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationAttributeTests.cs
stijnmoreels: Yes, of course.

src/Arcus.WebApi.Unit/Hosting/TestApiServer.cs
stijnmoreels: Not very much use to expose this to the rest of the test project; but no problem for me.

src/Arcus.WebApi.Security/Authentication/X509Validation.cs
stijnmoreels: Yes, it is indeed more a requirement. Good suggestion!

src/Arcus.WebApi.Unit/Hosting/TestApiServer.cs
tomkerkhove: It's to align with our other classes, thanks!

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
```diff
@@ -0,0 +1,71 @@
...
+        [InlineData("unrecognizedSubjectName", "known-issuername", true)]
+        [InlineData("known-subject", "unrecognizedIssuerName", true)]
+        [InlineData("unrecognizedSubjectName", "unrecognizedIssuerName", true)]
+        public async Task AuthorizedRoute_WithCertificateAuthentication_ShouldFailWithUnauthorized_WhenAnyClientCertificateValidationDoesntSucceeds(
+            string subjectName,
+            string issuerName,
+            bool expected)
+        {
+            // Arrange
+            _testServer.AddFilter(new CertificateAuthenticationFilter(X509ValidationRequirement.SubjectName, "CN=known-subject"));
```
tomkerkhove: I presume this is similar to how I would configure the global certificate enforcement, right?
Wouldn't it make more sense if I only have to add one of them which takes one or more requirements?

tomkerkhove: From a consumer perspective this would be nicer I think.

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationAttributeTests.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Arcus.WebApi.Unit.Hosting;
+using Arcus.WebApi.Unit.Security.Doubles;
+using Xunit;
+using static Arcus.WebApi.Unit.Security.Authentication.CertificateAuthenticationOnMethodController;
+
+namespace Arcus.WebApi.Unit.Security.Authentication
```
tomkerkhove: We should have a test for `Thumbprint` as well

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Initializes a new instance of the <see cref="CertificateAuthenticationFilter"/> class.
+        /// </summary>
+        /// <param name="requirement">The property of the client <see cref="X509Certificate2"/> to validate.</param>
+        /// <param name="expectedValue">The expected value the property of the <see cref="X509Certificate2"/> should have.</param>
+        public CertificateAuthenticationFilter(X509ValidationRequirement requirement, string expectedValue)
+        {
+            Guard.NotNullOrWhitespace(expectedValue, nameof(expectedValue), "Expected value in certificate cannot be blank");
+
+            _requirement = requirement;
+            _expectedValue = expectedValue;
```
tomkerkhove: I can see value for this, but we should use be more flexible than this. We don't want to restart because the value has changed.

So I would suggest providing a way where we can specify how to retrieve it via `IConfiguration`

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,94 @@
...
+using GuardNet;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+
+namespace Arcus.WebApi.Security.Authentication
+{
+    /// <summary>
+    /// Authentication filter to secure HTTP requests by allowing only certain values in the client certificate.
+    /// </summary>
+    public class CertificateAuthenticationFilter : IAuthorizationFilter
```
tomkerkhove: We should pull in `ILogger` in here and provide information traces of what is going on. This should help customers analyse issues.

Scenarios I'm thinking of where I'd like to have information:
1. No client cert was specified
2. X509ValidationRequirement.SubjectName failed and why (incl expected, actual)
2. X509ValidationRequirement.IssuerName failed and why (incl expected, actual)
2. X509ValidationRequirement.Thumbprint failed and why (incl expected, actual)

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = true)]
+    public class CertificateAuthenticationAttribute : TypeFilterAttribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CertificateAuthenticationAttribute"/> class.
+        /// </summary>
+        /// <param name="requirement">The property of the client <see cref="X509Certificate2"/> to validate.</param>
+        /// <param name="expectedValue">The expected value the property of the <see cref="X509Certificate2"/> should have.</param>
+        public CertificateAuthenticationAttribute(X509ValidationRequirement requirement, string expectedValue) : base(typeof(CertificateAuthenticationFilter))
```
tomkerkhove: `CertificateAuthenticationAttribute` should not depend on `CertificateAuthenticationFilter`, I see this more as both of these using a shared certificate authentication class as part of it.

tomkerkhove:
I have also noticed that our `ExceptionHandlingMiddleware` is using the following approach to enable it:
```csharp
public void Configure(IApplicationBuilder app, IHostingEnvironment env)
{

   ...
   app.UseMvc();
}
```

While the SharedAccessKeyAuthenticationFilter & this approach is the following:
```csharp
public void ConfigureServices(IServiceCollections services)
{
    services.AddScoped<ICachedSecretProvider>(serviceProvider => new MyCachedSecretProvider());
    services.AddMvc(options => options.Filters.Add(new SharedAccessKeyAuthenticationFilter(headerName: "http-request-header-name", secretName: "shared-access-key-name")));
}
```

Is the latter approach different given the authentication nature of it or what was the reason again? Please refresh my memory 😄 

/cc @fgheysels

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
stijnmoreels: Not sure what you mean with this. Structuring this way is the recommended way to do this.
It is only dependend on the type to create instances, not on a single (thus shared) instance.

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
stijnmoreels: Ah, I see your point. That could be done and would make the configuration more simpler.

fgheysels:
> I have also noticed that our `ExceptionHandlingMiddleware` is using the following approach to enable it:

> Is the latter approach different given the authentication nature of it or what was the reason again? Please refresh my memory 😄
> 
> /cc @fgheysels

There is a difference, and you can see it in the code already:
The filters operate within the MVC context whereas the middleware operates before the MVC thing kicks in.
The ExceptionHandling functionality has been implemented as middleware since it doesn't need to access routing data etc...

I think the Authentication functionality has been implemented as a filter because you might want to apply it on a specific set of operations or exclude certain operations.  Middleware will be executed on each request since it is registered in the asp.net pipeline.   The 'Filters' will only be executed on operations that are decorated with that attribute.
/correct-me-if-im-wrong :P

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
fgheysels: I think @tomkerkhove means 'favor composition over inheritance' ?

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
fgheysels: You mean to have something like this:

```csharp
new CertificateAuthenticationFilter(new CertificateRequirement(X509ValidationRequirement.SubjectName, "CN=.."),
                                    new CertificateRequirement(X509ValidationRequirement.IssuerName, "CN=kn..."));
```
?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
fgheysels: Wouldn't it be an idea to create a feature-request for this, and implement that in a separate pull-request ?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
tomkerkhove: That is correct, sorry for not being clear!

tomkerkhove:
That sound pretty clear, thanks for the update!

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
tomkerkhove: Personally I would expect to give one "configuration" object that defines how the validation should work, but maybe a list could be sufficient as well but an object feels more gentle.

Again, personal preference - How we do it is not that important for me, as long as it's nice and simple

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
tomkerkhove: Personally, I don't see it as a feature request given you will never add an attribute with a fixed thumbprint for example. It will always come from the configuration, no?

Does that mean we should not have that option? I don't think so.

But I get your point of doing seperate PR to simplify this one, tracking in #44

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
stijnmoreels: Hmm, still not clear to me. As I said, this way of defining both local and global access policies are default approaches because of the exact same implementation.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: Yes, will do; is this something that from the services should be retrieved?

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
stijnmoreels: Yes, I take a look at that.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: Yes, will do!

stijnmoreels:
Question @tomkerkhove : I worked locally on this PR-branch to pass keys to the filter/attribute instead of the actual value but I'm not sure if the configuration should be an `IConfiguration` instance from ASP.NET or our own `ISecretProvder`?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
tomkerkhove: That is correct, but one approach depends on the other where my suggestion is to move out the effective validation logic into a seperate class which is used by both local and global access policies.

Let's say a `ICertificateValidator` which is used by both `CertificateAuthenticationAttribute` and `CertificateAuthenticationFilter`

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
```diff
@@ -0,0 +1,129 @@
...
+            bool expected)
+        {
+            // Arrange
+            const string subjectKey = "subject", issuerKey = "issuer";
+            _testServer.AddService<ISecretProvider>(
+                new InMemorySecretProvider(
+                    (subjectKey, "CN=known-subject"),
+                    (issuerKey, "CN=known-issuername")));
+
+            _testServer.AddFilter(
```
tomkerkhove: This notation feels very unnatural for object-oriented programming, why are we not using a list or object with all requirements inside of it?

For example:
```csharp
var certificateRequirements = new CertificateRequirements
{
    SubjectNames = new List<string>() {subjectKey},
    IssuerNames= new List<string>() {issuerKey},
};

_testServer.AddFilter(new CertificateAuthenticationFilter(certificateRequirements));
```

What do you think @fgheysels

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,171 @@
...
+            X509Certificate2 clientCertificate = context.HttpContext.Connection.ClientCertificate;
+            if (clientCertificate == null)
+            {
+                logger.LogWarning(
+                    "No client certificate was specified in the HTTP request while this authentication filter "
+                    + $"requires a certificate to validate on the {String.Join(", ", _requirements.Select(item => item.requirement))}");
+                
+                context.Result = new UnauthorizedResult();
+            }
+            else if (!await IsAllowedCertificate(clientCertificate, userDefinedSecretProvider, logger))
```
tomkerkhove: Please don't inline this stuff :)

Basically, everything that is done here is what I was referring to where I would move that out of this class.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,171 @@
...
+            ISecretProvider provider,
+            ILogger logger)
+        {
+            var requirementValues = await Task.WhenAll(_requirements.Select(async item =>
+            {
+                string expected = await provider.Get(item.configurationKey);
+                return (requirement: item.requirement, key: item.configurationKey, expected: expected);
+            }));
+
+            return requirementValues.All(value =>
```
tomkerkhove: Move all of this logic to dedicated method instead?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,171 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Arcus.WebApi.Security.Authentication
+{
+    /// <summary>
+    /// Authentication filter to secure HTTP requests by allowing only certain values in the client certificate.
+    /// </summary>
+    public class CertificateAuthenticationFilter : IAsyncAuthorizationFilter
+    {
+        private readonly (X509ValidationRequirement requirement, string configurationKey)[] _requirements;
```
tomkerkhove: Why are we not creating an object for this?

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
tomkerkhove: Based on reading the code it looks like `X509ValidationRequirement.SubjectName` is a constant string but if I have a look at the constructor this is actually a requirement or?

The code is not clear on that perspective imo, sorry :/

tomkerkhove:
@stijnmoreels 	I think they should have access to both as some things can come from the configuration while you should be possible to use Key Vault as well.

docs/features/certificate-authentication.md
```diff
@@ -0,0 +1,75 @@
...
+
+After that, the `CertificateAuthenticationAttribute` attribute can be applied on the controllers, or if more fine-grained control is needed, on the operations that requires authentication:
+
+```csharp
+[ApiController]
+[CertificateAuthentication(X509CertificateRequirement.IssuerName, "key-to-certificate-issuer-name")]
+public class MyApiController : ControllerBase
+{
+    [HttpGet]
+    [Route("authz/certificate)]
```
tomkerkhove:
```suggestion
    [Route("authz/certificate")]
```

docs/features/certificate-authentication.md
```diff
@@ -0,0 +1,75 @@
...
+The authentication requires an `ICachedSecretProvider` or `ISecretProvider` dependency to be registered with the services container of the <span>ASP.NET</span> request pipeline. This is typically done in the `ConfigureServices` method of the `Startup` class.
+Once this is done, the `CertificateAuthenticationFilter` can be added to the filters that will be applied to all actions:
+
+```csharp
+public void ConfigureServices(IServiceCollections services)
+{
+    services.AddScoped<ICachedSecretProvider(serviceProvider => new MyCachedSecretProvider());
+    services.AddMvc(
+        options => options.Filters.Add(
+            new CertificateAuthenticationFilter(
```
tomkerkhove: Given this will always be the case, would it be convenient if we use `CertificateAuthenticationFilter.Create` as well?

docs/features/certificate-authentication.md
```diff
@@ -0,0 +1,75 @@
+---
```
tomkerkhove: Please add a link to the root of the docs and align the name of the file with shared-access-key.
Maybe consider moving it to /features/auth/<>

stijnmoreels: Yes, will do.

docs/features/certificate-authentication.md
stijnmoreels: Hmm, normally I would only create such a _Factory Method_ if the creation of the instance of the type requires more than just assignment. In this case there isn't any heavy-lifting done during creation so I don't see immediately any reason to also add such a creation mechanism?

stijnmoreels:
> @stijnmoreels I think they should have access to both as some things can come from the configuration while you should be possible to use Key Vault as well.

That you should explain :)

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
stijnmoreels: Ok, you could move the validation logic; but that doesn't change the fact that the attribute still requires a filter implementation to work properly; and will therefore still requires the certificate filter.
That's just how type filters work; so the only real change is yet another abtraction layer that is used only in a single class.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: Hmm, I'm not sure if that would improve readability; since I see enough relation between the two functions to place them toghether in a way seperate methods/names would make things maybe not that clear.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: Hmm, because it's short-lived, could have not so much added value as it comes to functionality, just data that only makes sense inside the certificate validation.

stijnmoreels: But I'm open to suggestions if you could provide an alternative.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: It is strange, isn't it? How this was only a problem when the result was changed from `bool` to `Task<bool>` and an additional `await` was required.
I would love to hear more on how you find this a reason to introduce an temporary variable.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
fgheysels: Actually, this is just a named Tuple.
It's in fact the same as 
```csharp
private readonly  Tuple<X509ValidationRequirement, string>[] _requirements;
```
The advantage with the code in the PR however, is that the properties of the tuple have a custom name (`requirement` and `configurationKey`) whereas when using the `Tuple<T1,T2>` class, this would be `Item1` and `Item2`.
Therefore, I think that using the named Tuple feature of C# is a better approach then using the `Tuple<T1,T2>` class and I also think that it would be overkill to create a custom type just for this.

One suggestion though:  instead of naming the properties `requirement` and `configurationKey`, I'd call them `Requirement` and `ConfigurationKey`.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,171 @@
...
+                
+                context.Result = new UnauthorizedResult();
+            }
+            else if (!await IsAllowedCertificate(clientCertificate, userDefinedSecretProvider, logger))
+            {
+                context.Result = new UnauthorizedResult();
+            }
+        }
+
+        private async Task<bool> IsAllowedCertificate(
```
fgheysels: I'd rename this method to `IsCertificateAllowedAsync`.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,171 @@
...
+            X509Certificate2 clientCertificate, 
+            ISecretProvider provider,
+            ILogger logger)
+        {
+            var requirementValues = await Task.WhenAll(_requirements.Select(async item =>
+            {
+                string expected = await provider.Get(item.configurationKey);
```
fgheysels: It would be better if all async methods are suffixed with `Async` IMO.  It makes it immediately obvious that the method is async and that you'd probably want to `await` it.  (But I believe this `ISecretProvider.Get` method comes from another arcus repo ?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
fgheysels: It's actually a matter of preference and readability imho.
In most cases, I'd also prefer to create a temp variable that holds the result instead of awaiting in the if condition.  In this case however, I think I'd leave it like it is since it would require you to rewrite it as:

```csharp
else
{
     var isCertAllowed = await IsCertificateAllowedAsync(....);
     if( !isCertAllowed )
     {
```

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
fgheysels: I think Tom suggests to move the inline lambda expression to a method, which I can agree with.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: All true, except that named tuples are actually `ValueTyle<_, _>` instead of `Tuple<_, _>` ;)

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: I think that too, but since all the Arcus async methods are not written this way; it would break consistency.

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
fgheysels: I wouldn't use a `List` type as a parameter for the constructor of the `CertificateRequirements` class since it is to concrete and has functionality that is not necessary.
The code in the PR is possible because of the `params` keyword in the constructor.  It allows you to write an array as a comma separated list of values.  Given the fact that the type of the array is a ValueTuple, the syntax `( ... ) , ( ... )` is possible.
It is true that it can be somewhat confusing when you read it.

Therefore, I would suggest to change the constructor of the `CertificateAuthenticationFilter` to:

```csharp
public CertificateAuthenticationFilter( IEnumerable<(X509ValidationRequirement, string)> )
```
So, using an `IEnumerable` instead of the more concrete/specific array type.

This means that the code in question needs to be written like this:

```csharp
new CertificateAuthenticationFilter( new [] 
                                     { 
                                         (X509ValidationRequirement.SubjectName, subjectKey),
                                         (X509ValidationRequirement.IssuerName, issuerKey)
                                     });
```

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
fgheysels: We could alias the value-tuple, [once it is supported](https://github.com/dotnet/csharplang/issues/423) :)

stijnmoreels: That would be cool. Yet another feature that's already supported in F# ;)

tomkerkhove: It does not bring value in terms of functionality but it's all about readability in my opinion. If you have a look at the consumer side it's not that pretty imo

docs/features/certificate-authentication.md
tomkerkhove: Sounds good to me. Resolving

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
tomkerkhove: I thought this was defined in the arcus coding guidlines which we defined back in the day but apparently not. https://codit.visualstudio.com/Arcus/_wiki/wikis/Arcus.wiki?wikiVersion=GBwikiMaster&pagePath=%2FGuidance%2FCoding%20Guidelines&pageId=10

I am on the same page on that perspective, however we need to see how our public methods are now and if we should align them.

Think a seperate issue for tracking this will be best option.

tomkerkhove:
> > @stijnmoreels I think they should have access to both as some things can come from the configuration while you should be possible to use Key Vault as well.
> 
> That you should explain :)

Basically, it's not up to us to enforce where the value should be stored. Consumers should have the capability to decide if it is coming from the configuration or not.

Let's sync about this offline

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
fgheysels: It is true that the ValueType approach is not that very obvious.   I tend to use it in private parts of classes; since it is here also exposed by the public ctor, it might indeed be a better option to create a dedicated type (which is not that of a big deal):

```csharp
public class CertificateRequirement
{
      public X509ValidationRequirement RequirementType { get; }
      public string RequirementValue { get; }
}
```

```csharp
public class CertificateAuthenticationFilter
{
     public CertificateAuthenticationFilter(IEnumerable<CertificateRequirement> requirements )
     {}
}
```

```csharp
new CertificateAuthenticationFilter(new [] 
                                    {
                                        new CertificateRequirement(X509ValidationRequirement.Subject, ""),
                                        new CertificateRequirement(X509ValidationRequirement.Thumbprint,  "")
                                    });
```

stijnmoreels:
- certificate validation sep
- certificate config single configuration
	> registered once
	> each cert property with own retrieval
	> attribute only prop/key

docs/features/auth/certificate-authentication.md
```diff
@@ -0,0 +1,99 @@
+---
+title: "Authentication with certificate via ASP.NET Core authentication filters"
+layout: default
+---
+
+The `Arcus.WebApi.Security` package provides a mechanism that uses the client certificate of the request to grant access to a web application.
+This authentication process consists of following parts:
```
tomkerkhove: Let's add a break before listing flow?
```suggestion

This authentication process consists of the following steps:
```

docs/index.md
```diff
@@ -30,7 +30,8 @@ PM > Install-Package Arcus.WebApi.Logging
 
 The `Arcus.WebApi.Security` package contains functionality to easily add security capabilities to an API.
 
-- [Shared access key authentication](features/shared-access-key.md)
```
tomkerkhove: Careful, I did some changes to this one recently

docs/index.md
```diff
@@ -30,7 +30,8 @@ PM > Install-Package Arcus.WebApi.Logging
...
+- [Shared access key authentication](features/auth/shared-access-key.md)
+- [Certificate authentication](features/auth/certificate-authentication.md)
```
tomkerkhove: I like this, but don't think we need the `authentication` suffix anymore
```suggestion
- [Certificate authentication](features/auth/certificate.md)
```

docs/features/auth/certificate-authentication.md
```diff
@@ -0,0 +1,99 @@
...
+
+Each certificate property that should be validated can use a different service dependency. 
+This mapping of what service which property uses, is defined in an `CertificateAuthenticationValidator` instance.
+
+Once this is done, the `CertificateAuthenticationFilter` can be added to the filters that will be applied to all actions:
+
+```csharp
+public void ConfigureServices(IServiceCollections services)
+{
+    services.AddScoped<ICachedSecretProvider(serviceProvider => new MyCachedSecretProvider());
```
tomkerkhove: I think this approach deviates a bit from what we've discussed.

My understanding is that we will have a `CertificateAuthenticationConfigurationBuilder` that allows us to build the config of what to validate, for example:
```csharp
var certificateAuthenticationConfig = CertificateAuthenticationValidator
                                          .WithSubject(X509InputStore.Configuration, "my-config-key") // X509InputStore is a terrible name
                                          .WithThumbprint(X509InputStore.SecretProvider, "my-secret-name")
                                          .Build()
```

Which is then passed to the Filter to be used to run your validation against with the following configuration/criteria
```csharp
services.AddMvc(options  =>
   options.Filters.Add(new CertificateAuthenticationFilter(certificateAuthenticationConfig));
```

`CertificateAuthenticationRequirementsBuilder` is also fine for me and maybe is more explicit but the idea is the outcome of the builder is a single object, for example `CertificateAuthenticationRequirement`, which contains everything needed by the validator to run and know where to pull info from.

Did you have the same understanding @fgheysels?

docs/features/auth/certificate-authentication.md
```diff
@@ -0,0 +1,99 @@
...
+- `ICachedSecretProvider` or `ISecretProvider`: built-in or you implementation of the secret provider
+- `Configuration`: key/value pairs in the configuration of the <span>ASP.NET</span> application
+- `IX509ValidationLocation`: custom implementation that implements the validation location interface
+
+This registration of the service is typically done in the `ConfigureServices` method of the `Startup` class:
+
+```csharp
+public void ConfigureServices(IServiceCollections services)
+{
+    services.AddScoped<ICachedSecretProvider(serviceProvider => new MyCachedSecretProvider());
```
tomkerkhove: Here you would have to do the same setup as mentioned above and then in the controller itself you can decorate it simply with `[CertificateAuthentication)]`.

If people want more flexibility *later on*, we can track this in the backlog to create some sort of profiles, but this is too far down the road for now.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            _requirements = requirements.ToDictionary(requirement => requirement.Key, requirement => new ConfiguredKey(requirement.Value));
+        }
+
+        /// <summary>
+        /// Called early in the filter pipeline to confirm request is authorized.
+        /// </summary>
+        /// <param name="context">The <see cref="T:Microsoft.AspNetCore.Mvc.Filters.AuthorizationFilterContext" />.</param>
+        public async Task OnAuthorizationAsync(AuthorizationFilterContext context)
+        {
```
tomkerkhove: I like this a lot more - We do the orchestration but the validator does all the magic, love it!

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
```diff
@@ -0,0 +1,85 @@
...
+            Guard.NotNull(context, nameof(context));
+            Guard.NotNull(context.HttpContext, nameof(context.HttpContext));
+            Guard.For<ArgumentException>(() => context.HttpContext.Connection is null, "Invalid action context given without any HTTP connection");
+            Guard.For<ArgumentException>(() => context.HttpContext.RequestServices is null, "Invalid action context given without any HTTP request services");
+
+            X509Certificate2 clientCertificate = context.HttpContext.Connection.ClientCertificate;
+            if (clientCertificate == null)
+            {
+                ILogger logger =
```
tomkerkhove: Maybe move the logger determination to a small method?
```suggestion
                ILogger logger = GetLogger()
```

When reading this method I don't really care where the logger comes from, but this is just my opinion.
You don't have to change it if you don't agree.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
```diff
@@ -0,0 +1,252 @@
...
+
+            IDictionary<X509ValidationRequirement, (ConfiguredKey configuredKey, string expected)> requirementValues = 
+                await GetExpectedCertificateValues(requirements, services);
+            
+            ILogger logger = 
+                services.GetService<ILoggerFactory>()
+                       ?.CreateLogger<CertificateAuthenticationValidator>() 
+                ?? (ILogger) NullLogger.Instance;
+
+            return requirementValues.All(keyValue =>
```
tomkerkhove: Same feedback, move lambda to a dedicated method which is called instead?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
```diff
@@ -0,0 +1,252 @@
...
+            Guard.NotNull(requirements, nameof(requirements), "Certificate authentication validation requires a series of requirements and their configured key");
+            Guard.NotNull(services, nameof(services), "Certificate authentication validation requires a service object to retrieve registered services");
+            Guard.For<ArgumentException>(
+                () => requirements.Any(requirement => requirement.Value is null), 
+                "Certificate authentication requires all configured keys in the series of requirements to be not 'null'");
+
+            IDictionary<X509ValidationRequirement, (ConfiguredKey configuredKey, string expected)> requirementValues = 
+                await GetExpectedCertificateValues(requirements, services);
+            
+            ILogger logger =
```
tomkerkhove: Same feedback, do logger determination here or move to small method?

docs/features/certificate-authentication.md
tomkerkhove: Done, thanks!

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
tomkerkhove: Discussed to use named tuples only for internals which is fine, resolving

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
tomkerkhove: This has been changed and I like it a lot!

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
tomkerkhove: See most latest comment on new builder approach.

Resolving this in favor of the other one.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
tomkerkhove: Validation logic was moved to dedicated class, resolving

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
tomkerkhove: Created https://github.com/arcus-azure/arcus/issues/28 for this

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
tomkerkhove: Created https://github.com/arcus-azure/arcus/issues/28 for this

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationAttributeTests.cs
stijnmoreels: That's done I think.

docs/features/auth/certificate-authentication.md
stijnmoreels: Yep.

docs/index.md
stijnmoreels: Ok, just merged with remote upstream. No conflicts.

docs/index.md
stijnmoreels: Yep, ok will do.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: :+1:

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
stijnmoreels: Let me check

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
stijnmoreels: I restructured some of the methods so it requires less arguments and is more simpler.

docs/features/auth/certificate-authentication.md
stijnmoreels: Ah, yes; now I understand it more. I didn't was quite a fan of how it is now set-up because the configuration is splintered across the application.

stijnmoreels: But then we would have a fluent member `.With...` for each enum member. Isn't that what duplication? The alternative is passing allong, but that wouldn't probably make thinks easier.

docs/features/auth/certificate-authentication.md
tomkerkhove: Looks good!

docs/index.md
tomkerkhove: Awesome, thanks!

docs/features/auth/certificate-authentication.md
tomkerkhove: We would indeed have a `.With` for every enum value, but the consumers will not even know about that enumeration value given they will only use the builder or config directly.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationFilter.cs
tomkerkhove: Looks good now, thanks.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
```diff
@@ -0,0 +1,126 @@
...
+                        return IsCertificateThumbprintAllowed(clientCertificate, keyValue.Value, logger);
+                    default:
+                        throw new ArgumentOutOfRangeException(nameof(keyValue.Key), keyValue.Key, "Unknown validation type specified");
+                }
+            });
+        }
+
+        private static ILogger GetLoggerOrDefault(IServiceProvider services)
+        {
+            return services.GetService<ILoggerFactory>()
```
tomkerkhove: Would be more readable to capture output of Create Logger and do the short hand return if not null on a dedicated line.

This whole method is one statement which is an anti-pattern imo

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
tomkerkhove: THanks!

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
stijnmoreels: Not sure if I share that opinion of it being an anti-pattern since it allows devs to more easily add null-checks. Which I would consider a plus.

You can argue of course about readability and such; certainly if we want to express our Arcus code-base in the most basic possible way to not lose beginner devs in syntax.

tomkerkhove: I was more referring to this as it's more readable, if you all do it on one line it's not really reable
```csharp
ILogger logger = services.GetService<ILoggerFactory>()
                         ?.CreateLogger<CertificateAuthenticationFilter>();

return logger ?? NullLogger.Instance;
```

tomkerkhove: But current approach is ok as well, no worries

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
stijnmoreels: Outdated...

docs/features/auth/certificate-authentication.md
stijnmoreels: Outdated...

docs/features/auth/certificate-authentication.md
stijnmoreels: Yep, is now current approach.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
tomkerkhove: Not really - https://github.com/arcus-azure/arcus.webapi/pull/43/files#diff-a00c4d616f7db7b54040a476294f8c9dR52

stijnmoreels: Ah, OK, I see.
Not sure if that would make it more readable? Now you see that every certificate requirement goes through the dedicated kind of validation in the exposed (internal) method.

tomkerkhove: The reason why it would improve readability is that in the context of this method I don't care how the validation itself works, it just has to run it.

If I want to see the details, then I go to the method itself. 

In the main method it would just look like the following:
`return expectedValuesByRequirement.All(PerformValidation)`

Next to that, the list currently is fairly short but the more checks we add, the longer this inline method call will be.

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationAttributeTests.cs
tomkerkhove: Thanks!

src/Arcus.WebApi.Security/Authentication/X509ValidationRequirement.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Security.Cryptography.X509Certificates;
+
+namespace Arcus.WebApi.Security.Authentication 
+{
```
tomkerkhove: Nice docs, thank you!

src/Arcus.WebApi.Security/Authentication/X509ValidationLocation.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Security.Cryptography.X509Certificates;
+using Arcus.Security.Secrets.Core.Interfaces;
+using Microsoft.Extensions.Configuration;
+
+namespace Arcus.WebApi.Security.Authentication
+{
```
tomkerkhove: Same here, nice docs - Thanks!

src/Arcus.WebApi.Security/Authentication/SecretProviderValidationLocation.cs
```diff
@@ -0,0 +1,49 @@
...
+                ?? services.GetService<ISecretProvider>();
+
+            if (userDefinedSecretProvider == null)
+            {
+                throw new KeyNotFoundException(
+                    $"No configured {nameof(ICachedSecretProvider)} or {nameof(ISecretProvider)} implementation found in the request service container. "
+                    + "Please configure such an implementation (ex. in the Startup) of your application");
+            }
+
+            Task<string> getValueAsync = userDefinedSecretProvider.Get(configurationKey);
```
tomkerkhove: Just curious but why don't we call the async method and run the logic on that?

src/Arcus.WebApi.Security/Authentication/ConfigurationValidationLocation.cs
```diff
@@ -0,0 +1,53 @@
...
+                        $"No configured {nameof(IConfiguration)} implementation found in the request service container. "
+                        + "Please configure such an implementation (ex. in the Startup) of your application");
+                }
+
+                string value = configuration[configurationKey];
+                return Task.FromResult(value);
+            }
+            catch (Exception ex)
+            {
+                return Task.FromException<string>(ex);
```
tomkerkhove: TIL - I did not know that existed, thanks.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
```diff
@@ -0,0 +1,133 @@
...
+                return logger;
+            }
+
+            return NullLogger.Instance;
+        }
+
+        private static bool IsCertificateSubjectNameAllowed(X509Certificate2 clientCertificate, ExpectedCertificateValue expected, ILogger logger)
+        {
+            IEnumerable<string> certificateSubjectNames =
+                (clientCertificate.Subject ?? String.Empty)
```
tomkerkhove: If the subject is empty, do we even need to go through the whole process?

Maybe an early return would be better

tomkerkhove: Nah it's ok.

docs/features/auth/certificate.md
```diff
@@ -0,0 +1,98 @@
...
+- `IX509ValidationLocation`/`X509ValidationLocation`: custom or built-in implementation that retrieves the expected certificate values
+
+This registration of the service is typically done in the `ConfigureServices` method of the `Startup` class:
+
+```csharp
+public void ConfigureServices(IServiceCollections services)
+{
+    services.AddScoped<ICachedSecretProvider(serviceProvider => new MyCachedSecretProvider());
+
+    services.AddScoped<CertificateAuthenticationValidator>(
```
tomkerkhove: For sake of documentation I would improve it as following:
```csharp
var certificateAuthenticationConfig = new CertificateAuthenticationConfig()
                                            .WithIssuer(X509ValidationLocation.SecretProvider, "key-to-certificate-issuer-name");
 var certificateAuthenticationValidator = new CertificateAuthenticationValidator(certificateAuthenticationConfig);

services.AddScoped<CertificateAuthenticationValidator>(serviceProvider => certificateAuthenticationValidator);
```

Or even:
```csharp
var certificateAuthenticationConfig = new CertificateAuthenticationConfig()
                                            .WithIssuer(X509ValidationLocation.SecretProvider, "key-to-certificate-issuer-name");

services.AddScoped<CertificateAuthenticationConfig>(serviceProvider => certificateAuthenticationConfig);
services.AddScoped<CertificateAuthenticationValidator>();
```

docs/features/auth/certificate.md
```diff
@@ -0,0 +1,98 @@
...
+This mapping of what service which property uses, is defined in an `CertificateAuthenticationValidator` instance.
+
+Once this is done, the `CertificateAuthenticationFilter` can be added to the filters that will be applied to all actions:
+
+```csharp
+public void ConfigureServices(IServiceCollections services)
+{
+    services.AddScoped<ICachedSecretProvider(serviceProvider => new MyCachedSecretProvider());
+
+    services.AddScoped<CertificateAuthenticationValidator>(
```
tomkerkhove: Same here, simplify.

docs/features/auth/certificate.md
```diff
@@ -0,0 +1,98 @@
+---
```
tomkerkhove: Given the configuration of the attributes are the same, but the way they are applied not I would maybe reshuffle the docs and promote the configurate to a dedicated section and point to it from the 2 attributes/filters.

Would that make sense?

src/Arcus.WebApi.Security/Authentication/ConfiguredKey.cs
```diff
@@ -0,0 +1,35 @@
+﻿using GuardNet;
+
+namespace Arcus.WebApi.Security.Authentication
+{
+    /// <summary>
+    /// Represents the configured key on the authentication mechanisms.
+    /// </summary>
+    internal class ConfiguredKey
```
tomkerkhove: Just checking, where is this being used?

tomkerkhove: Ok I've found it but I'm just curious why we actually need it - We guard for null over there so this might be overkill or am I missing something?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationConfig.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        /// <summary>
+        /// Configures the validation for the <see cref="X509Certificate2.Thumbprint"/> from a given <paramref name="location"/> using a specified <paramref name="configuredKey"/>.
+        /// </summary>
+        /// <param name="location">The location to retrieve the expected certificate thumbprint.</param>
+        /// <param name="configuredKey">The configured key that the <paramref name="location"/> requires to retrieve the expected thumbprint.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="location"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="configuredKey"/> is blank.</exception>
+        public CertificateAuthenticationConfig WithThumbprint(IX509ValidationLocation location, string configuredKey)
```
tomkerkhove: Is this to provide extensibility or?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationConfig.cs
```diff
@@ -0,0 +1,172 @@
...
+    /// Representation of the configurable validation requirements on a <see cref="X509Certificate2"/>.
+    /// </summary>
+    public class CertificateAuthenticationConfig
+    {
+        private readonly IDictionary<X509ValidationRequirement, (IX509ValidationLocation location, ConfiguredKey configuredKey)> _locationAndKeyByRequirement;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CertificateAuthenticationConfig"/> class.
+        /// </summary>
+        public CertificateAuthenticationConfig()
```
tomkerkhove: Since it's a builder pattern this one should not be public and kick off the builder process via a Builder class

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationConfig.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <param name="logger">The logger used during the retrieval of each expected certificate value.</param>
+        /// <returns>The key/value pair of which certificate requirement to validate together with which value in the actual client certificate to expect.</returns>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="services"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> is <c>null</c>.</exception>
+        internal async Task<IDictionary<X509ValidationRequirement, ExpectedCertificateValue>> GetAllExpectedCertificateValues(IServiceProvider services, ILogger logger)
+        {
+            Guard.NotNull(services, nameof(services), "Request services cannot be 'null'");
+            Guard.NotNull(logger, nameof(logger), "Logger cannot be 'null'");
+
+            var expectedValuesByRequirement = await Task.WhenAll(_locationAndKeyByRequirement.Select(async keyValue =>
```
tomkerkhove: This is another case where all the logic in the select should be moved to a method instead

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
```diff
@@ -0,0 +1,230 @@
...
+    {
+        private readonly TestApiServer _testServer = new TestApiServer();
+
+        [Fact]
+        public async Task AuthorizedRoute_WithCertificateAuthentication_ShouldFailWithUnauthorized_WhenClientCertificateSubjectNameDoesntMatch()
+        {
+            // Arrange
+            string subjectKey = "subject", subjectValue = $"subject-{Guid.NewGuid()}";
+            _testServer.AddService<ISecretProvider>(new InMemorySecretProvider((subjectKey, subjectValue)));
+            _testServer.AddService(
```
tomkerkhove: Just a general remark for all tests, if you register CertificateAuthenticationValidator seperately from CertificateAuthenticationConfig you can avoid calling it like this and switch to :
```csharp
 _testServer.AddService(CertificateAuthenticationConfig()
                            .WithSubject(X509ValidationLocation.SecretProvider, subjectKey));
```

And move the registration of `CertificateAuthenticationValidator` to the test server

tomkerkhove: Just a suggestion to reduce duplication

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationValidator.cs
tomkerkhove: Thanks!

tomkerkhove:
One more thing - In terms of builder-pattern, the object is only created after the first call or at the end of it.

Currently we use the following approach:
```csharp
new CertificateAuthenticationConfig()
     .WithSubject(X509ValidationLocation.SecretProvider, subjectKey)
     .WithIssuer(X509ValidationLocation.SecretProvider, issuerKey)
```

I'd suggest to go to a similar approach to align with the usual pattern:
```csharp
CertificateAuthenticationConfigBuilder()
     .WithSubject(X509ValidationLocation.SecretProvider, subjectKey)
     .WithIssuer(X509ValidationLocation.SecretProvider, issuerKey)
     .Build() // Create CertificateAuthenticationConfig here
```

src/Arcus.WebApi.Security/Authentication/SecretProviderValidationLocation.cs
stijnmoreels: Aha, good you ask. It's because the interface is available for other users to implement; and since it returns a `Task` class, we could get `null` as return value; which would result in a `NullRefExc...` when we try to await on that.

```csharp
public Task<string> GetExpectedCertificateValueForConfiguredKey(blabla)
{
    return null;
{
```

_Trust no one_

docs/features/auth/certificate.md
stijnmoreels: :+1:

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationConfig.cs
stijnmoreels: Yep, to implement own implementation.

src/Arcus.WebApi.Security/Authentication/ConfiguredKey.cs
stijnmoreels: I added this both for readability (instead of using `string`'s all over and not knowing what it is + securing the use of the configured key to only verify this once and centrilized.

docs/features/auth/certificate.md
stijnmoreels: Not exactly, '2 attributes'? The registration of the validator is used in both sections; but otherwise it's different?

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationConfig.cs
stijnmoreels: Not sure what you mean; I think anything added above (adding yet another class) this configuration should break this "config" abstraction.

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
stijnmoreels: > And move the registration of CertificateAuthenticationValidator to the test server

That would not be a good idea if we want to reuse a clean-slate test server for our tests.

The registration  of the validator is not the only duplication that exists because of the 'simple'-testcase setup we must use and if we would follow allong this 'basic-approach' it would make sense to explicitly register it directly in the test case and not somewhere hidden.

stijnmoreels:
> One more thing - In terms of builder-pattern, the object is only created after the first call or at the end of it.
> 
> Currently we use the following approach:
> 
> ```cs
> new CertificateAuthenticationConfig()
>      .WithSubject(X509ValidationLocation.SecretProvider, subjectKey)
>      .WithIssuer(X509ValidationLocation.SecretProvider, issuerKey)
> ```
> 
> I'd suggest to go to a similar approach to align with the usual pattern:
> 
> ```cs
> CertificateAuthenticationConfigBuilder()
>      .WithSubject(X509ValidationLocation.SecretProvider, subjectKey)
>      .WithIssuer(X509ValidationLocation.SecretProvider, issuerKey)
>      .Build() // Create CertificateAuthenticationConfig here
> ```
> 
> Thoughts?

It would be more in line with the syntax of the 'textbook' builder-pattern; but since the `.Build()` would only pass allong the created config dictionary to another class; I don't think there is any 'building logic' left.

Mayybe it's me, but I always try to come-up with reasonable creation approaches with sometimes the use of patterns but not litterly duplication of patterns.

If you say that we should also here use the 'basic-approach'; than OK, I'll change this.

docs/features/auth/certificate.md
tomkerkhove: You are right, sorry about that!

src/Arcus.WebApi.Security/Authentication/SecretProviderValidationLocation.cs
tomkerkhove: 👌 thinking

src/Arcus.WebApi.Security/Authentication/ConfiguredKey.cs
tomkerkhove: 🙆‍♂

src/Arcus.WebApi.Unit/Security/Authentication/CertificateAuthenticationFilterTests.cs
tomkerkhove: I don't really see a problem why we could not do it but is ok, was just a suggestion.

Resolving!

tomkerkhove:
The reason I'd prefer a builder here is that we can create a new instance which will not contain any configuration, which is ok, but that might be done by accident.

For me this would be more explicit:
```csharp
var config = CertificateAuthenticationConfigBuilder()
               .WithSubject(X509ValidationLocation.SecretProvider, subjectKey)
               .WithIssuer(X509ValidationLocation.SecretProvider, issuerKey)
               .Build();
```

Or to create an empty one:
```csharp
var config = CertificateAuthenticationConfigBuilder()
               .Build();
```

Anyhow, don't have a strong opinion, just figured it would align with the other frameworks in .NET. 

Pulling in @fgheysels 	to get quorum on this.

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationConfig.cs
tomkerkhove: It's already being discussed in the general thread so I'm resolving it.

Was referring to this:

```csharp
var config = CertificateAuthenticationConfigBuilder()
               .WithSubject(X509ValidationLocation.SecretProvider, subjectKey)
               .WithIssuer(X509ValidationLocation.SecretProvider, issuerKey)
               .Build();
```

docs/features/auth/certificate.md
tomkerkhove: Thanks man!

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationAttribute.cs
```diff
@@ -6,6 +6,9 @@ namespace Arcus.WebApi.Security.Authentication
     /// <summary>
     /// Authentication filter to secure HTTP requests by allowing only certain values in the client certificate.
     /// </summary>
+    /// <remarks>
+    ///     Please make sure you register an <see cref="CertificateAuthenticationValidator"/> instance in the request services container (ex. in the Startup).
```
tomkerkhove: Good point @stijnmoreels 	👍

src/Arcus.WebApi.Security/Authentication/CertificateAuthenticationConfig.cs
tomkerkhove: Thanks for moving @stijnmoreels 	!

tomkerkhove:
Thank you very much @stijnmoreels 	🙌

fgheysels:
> The reason I'd prefer a builder here is that we can create a new instance which will not contain any configuration, which is ok, but that might be done by accident.
> 
> For me this would be more explicit:
> 
> ```cs
> var config = CertificateAuthenticationConfigBuilder()
>                .WithSubject(X509ValidationLocation.SecretProvider, subjectKey)
>                .WithIssuer(X509ValidationLocation.SecretProvider, issuerKey)
>                .Build();
> ```
> 
> Or to create an empty one:
> 
> ```cs
> var config = CertificateAuthenticationConfigBuilder()
>                .Build();
> ```
> 
> Anyhow, don't have a strong opinion, just figured it would align with the other frameworks in .NET.
> 
> Pulling in @fgheysels to get quorum on this.

I would agree with Tom's opinion, since it's an explicit way of creating an object, and the object is only created when you call `Build`.

With the approach that is currently in the PR:
```csharp
new CertificateAuthenticationConfig()
     .WithSubject(X509ValidationLocation.SecretProvider, subjectKey)
     .WithIssuer(X509ValidationLocation.SecretProvider, issuerKey)
```
you're creating an object and changing it's state, and you allow others to change its state as well,  whereas with the explicit builder pattern, you can make the `CertificateAuthenticationConfig` class immutable I think.

tomkerkhove:
Agreed! We've switched to a builder now, only exception is that you need to create an instance of it which is ok.

So now you can build config which are fully ready. Only thing I didn't check was if the config constructor was internal.

tomkerkhove:
It's internal, all is well!

Issue #45, opened 2019-06-20
OAuth authorization integration when using SwaggerUI in an API


fgheysels:
It's always a hassle when you want to test/execute an API operation using SwaggerUI when the API operation requires a Jwt bearer token.
It would be helpfull if there is an ability that allows you to call the authority server that issues a token from Swagger UI and use that token to invoke the API operations from within SwaggerUI.

**Describe the solution you'd like**
This can be implemented by creating a custom `Swashbuckle.IOperationFilter`.

fgheysels:
As I see it, this feature does not belong in any of the existing packages in arcus.webapi, so I assume we'll need to create a new one.
Any suggestions on the name ?  arcus.webapi.swagger ? arcus.webapi.swaggerextensions ? arcus.webapi.swagger.extensions ? ... ?

tomkerkhove:
This certainly sounds like an interesting one, thanks for the suggestion!

Given OpenAPI is around and this is mainly just for Swashbuckle I'm considering `arcus.webapi.openapi.extensions.swashbuckle` or `arcus.webapi.openapi.extensions`. Thoughts?

fgheysels:
> Given OpenAPI is around and this is mainly just for Swashbuckle I'm considering `arcus.webapi.openapi.extensions.swashbuckle` or `arcus.webapi.openapi.extensions`. Thoughts?

I'd prefer the shortest one (`arcus.webapi.openapi.extensions`).
 
> Do you by any chance have a snippet that elaborates on how it works?

It would come down to, instead of having a line like this in your ConfigureServices method to enable Swagger for your API:
```csharp
services.AddSwaggerGen(c =>
            {
                c.SwaggerDoc("v1", new Info { Title = "My Api", Version = "v1" });
            });
```
You'd introduce it like this:
```csharp
services.AddSwaggerGenWithOAuthEnabledClient(
                swaggerDocName: "v1",
                apiMetaData: new Info { Title = "My Api", Version = "v1" },
                authorityUrl: "http://myTokenProvider",
                scopes: new Dictionary<string, string>
                        {
                            { "scope1", "Scope description" }
                        });
```
(I'd like to find a better name for the `AddSwaggerGenWithOAuthEnabledClient` method however :) ).

This would make sure that the Swagger UI page has an Authorize button and when clicking upon that button, an access token is requested from the authority (token issuer) for the selected scopes.

tomkerkhove:
[...]

PR #46, opened 2019-06-25
Adding Arcus.WebApi.OpenApi.Extensions project

fgheysels:
This PR introduces the OpenApi.Extensions project within Arcus.

The OpenApi.Extensions project contains one class (`OAuthAuthorizeOperationFilter`) that allows you to easily integrate oauth security definitions in your openapi documentation.

When you add the `OAuthAuthorizeOperationFilter` to your `SwaggerGen` call, the methods that require authorization get some extra open-api documentation (additional responses and security information) injected into the swagger.json file.

@stijnmoreels : any thoughts on how we could provide some unit-tests for this ?  Fact is that this attribute is used by swashbuckle to generate the swagger.json documentation file.   To me, it seems overkill to test it by creating an api, generate open-api documentation and verify if the generated JSON file contains the expected properties.

#45


src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OAuthAuthorizeOperationFilter"/> class.
+        /// </summary>
+        public OAuthAuthorizeOperationFilter(IEnumerable<string> scopes)
+        {
+            _scopes = scopes;
+        }
+
+        public void Apply(Operation operation, OperationFilterContext context)
```
stijnmoreels: I would also verify the incoming parameters.
`Guard` them both and each property that you're using afterwards for `null`.

docs/features/openapi.securitydefinitions.md
```diff
@@ -0,0 +1,31 @@
...
+   c.SwaggerDoc("v1", new Info { Title = "My API v1", Version = "v1" });
+
+   c.AddSecurityDefinition("oauth2", new OAuth2Scheme
+   {
+      Flow = "implicit",
+      AuthorizationUrl = $"{authorityUrl}connect/authorize",
+      Scopes = scopes
+   });
+
+   c.OperationFilter<AuthorizeCheckOperationFilter>("myApiScope1", "myApiScope2");
```
stijnmoreels: I suspect that you should use our operation filter here instead of the `AuthorizeCheckOperationFilter` from a [demo project](https://github.com/dvanherten/swagger-identityserver/blob/master/WebApi/AuthorizeCheckOperationFilter.cs). :)

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            var hasAuthorize = context.MethodInfo.GetCustomAttributes(true).OfType<AuthorizeAttribute>().Any() ||
```
stijnmoreels: Question: now only the built-in authorization attribute is used. Could that be extended with the security attributes that are also part of this project? (shared access key, ...)?

stijnmoreels: Also, the `DeclaringType` could be `null` and should be checked.

I also think that using the `.GetCustomAttribute<>` would be more useful here.
This returns `null` when no such attribute can be found.

And maybe (not blocking) the `hasAuthorize` should be  split in two booleans: `operationHasAuthorize` and `typeHasAuthorize` or some sorts.

stijnmoreels:
Thanks for the feature AND the complete project file (with docs and NuGet generation) AND provided documentation! Nice job.

docs/features/openapi.securitydefinitions.md
```diff
@@ -0,0 +1,31 @@
...
+```csharp
+services.AddSwaggerGen(c =>
+{
+   c.SwaggerDoc("v1", new Info { Title = "My API v1", Version = "v1" });
+
+   c.AddSecurityDefinition("oauth2", new OAuth2Scheme
+   {
+      Flow = "implicit",
+      AuthorizationUrl = $"{authorityUrl}connect/authorize",
+      Scopes = scopes
```
stijnmoreels: Maybe an example on the `scopes`?

stijnmoreels:
> @stijnmoreels : any thoughts on how we could provide some unit-tests for this ? Fact is that this attribute is used by swashbuckle to generate the swagger.json documentation file. To me, it seems overkill to test it by creating an api, generate open-api documentation and verify if the generated JSON file contains the expected properties.

The test project contains a `TestApiServer` where you could add the swagger configuration. This test project can also contain API controllers which in your case should have an `[Authorize]` attribute. I'm not sure what the expected response should be when calling swagger endpoints, but after the `Startup` is correctly configured using swagger, you can call your endpoints like you would in a "real" API application.

docs/features/openapi.securitydefinitions.md
tomkerkhove: What do you mean with "our operation filter"? Are you referring to the shared access signature + certificate filter?

If so, we've discussed this and will exclude it in the first iteration.

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
tomkerkhove: We have discussed this and will exclude it in the first iteration and see how it goes, we will probably provide something on top of this one to work with our own ones. We don't want people to force the dependency if they don't need it.

docs/features/openapi.securitydefinitions.md
```diff
@@ -0,0 +1,31 @@
+---
+title: "Adding security information to OpenAPI documentation"
+layout: default
+---
+
+## Adding OAuth security definition to API operations
+
+### Introduction
+
+When an API is secured via OAuth, it is helpful if the Open API documentation makes this clear via a security scheme and the API operations that require authorization automatically inform the consumer that it is possible that a 401 (Unauthorized) or 409 (Forbidden) response is returned.
```
tomkerkhove:
```suggestion
When an API is secured via OAuth, it is helpful if the Open API documentation makes this clear via a security scheme and the API operations that require authorization automatically inform the consumer that it is possible that a 401 Unauthorized or 403 Forbidden response is returned.
```

docs/features/openapi.securitydefinitions.md
```diff
@@ -0,0 +1,31 @@
...
+The `OAuthAuthorizeOperationFilter` that is part of this package exposes this functionality.
```
tomkerkhove: I'd go with something along those lines:
```suggestion
`OAuthAuthorizeOperationFilter` will enrich your OpenAPI documentation with the relevant information.
```

docs/features/openapi.securitydefinitions.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Usage
+
+To indicate that an API is protected by OAuth, and enable Swagger UI to retrieve a JWT token when invoking API operations via Swagger, add the following code to the `ConfigureServices` method of the `Startup` class:
```
tomkerkhove:
```suggestion
To indicate that an API is protected by OAuth, you need to add `AuthorizeCheckOperationFilter` as an `OperationFilter` when configuring Swashbuckles Swagger generation:
```

tomkerkhove: I would move the following more to the introduction:
> and enable Swagger UI to retrieve a JWT token when invoking API operations via Swagger

docs/features/openapi.securitydefinitions.md
```diff
@@ -0,0 +1,31 @@
...
+
+```csharp
+services.AddSwaggerGen(c =>
```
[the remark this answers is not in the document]
tomkerkhove: This is documentation so not a blocker for me, but think indeed that it would improve readability

docs/features/openapi.securitydefinitions.md
tomkerkhove: Actually it's passed to the filter already but not used here so they need to be aligned indeed

docs/index.md
```diff
@@ -24,6 +24,12 @@ To install the Arcus.WebApi.Logging package:
 PM > Install-Package Arcus.WebApi.Logging
 ```
```
tomkerkhove: I'm wondering if we actually need to start listing all of these packages or move it to the more specific docs? But that's something for a separate PR I guess

docs/index.md
```diff
@@ -38,6 +44,12 @@ The `Arcus.WebApi.Logging` package contains functionality that can be incorporat
 
 - [Logging unhandled exceptions](features/logging.md)
 
+## Arcus.WebApi.OpenApi.Extensions
```
tomkerkhove: Let's go with the following to make it more readable
```suggestion
## OpenAPI Extensions
```

We should do the same for the other headers imo

docs/index.md
```diff
@@ -38,6 +44,12 @@ The `Arcus.WebApi.Logging` package contains functionality that can be incorporat
...
+
+The `Arcus.WebApi.OpenApi.Extensions` package contains functionality that can be used to easily improve the Open API documentation of an API when making using of Swashbuckle.
+
+- [Exposing security information in Swashbuckle documentation](features/openapi.securitydefinitions.md)
```
tomkerkhove: Let's use a dash here for the sake of the URL
```suggestion
- [Exposing security information in Swashbuckle documentation](features/openapi-securitydefinitions.md)
```

It feels a bit odd, but that's my opinion:
image

Maybe `features/openapi/extensions` or `features/openapi/security-definitions` is even better?

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
```diff
@@ -0,0 +1,56 @@
...
+                               (
+                                   context.MethodInfo.DeclaringType.GetCustomAttributes(true).OfType<AuthorizeAttribute>().Any() &&
+                                   context.MethodInfo.GetCustomAttributes(false).OfType<AllowAnonymousAttribute>().Any() == false
+                               );
+
+            if (hasAuthorize)
+            {
+                if (operation.Responses.ContainsKey("401") == false)
```
tomkerkhove: This is interesting, and maybe something for more down the road, but do we add both or make it configurable?

docs/features/openapi.securitydefinitions.md
stijnmoreels: The operation filter in the PR :)

tomkerkhove: Ah I see, naming is not inline, sorry about that :)

fgheysels: Should be fixed.

docs/features/openapi.securitydefinitions.md
```diff
@@ -0,0 +1,31 @@
...
+   c.SwaggerDoc("v1", new Info { Title = "My API v1", Version = "v1" });
+
+   c.AddSecurityDefinition("oauth2", new OAuth2Scheme
+   {
+      Flow = "implicit",
+      AuthorizationUrl = $"{authorityUrl}connect/authorize",
+      Scopes = scopes
+   });
+
+   c.OperationFilter<OAuthAuthorizeOperationFilter>("myApiScope1", "myApiScope2");
```
stijnmoreels: These scopes are passed allong the filter in a way you would think the parameters are decorated with a `params` syntactic sugar; but the input is an `IEnumerable<_>`; not sure though if this is correct?

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
fgheysels: I could do that, but I wonder if it adds any value here.   This operation is being called by Swashbuckle and we do not have any control over this.  I guess we should rely on the stability of Swashbuckle that the operation gets called with the correct parameters in this case ?

tomkerkhove: I tend to agree, if it gives null it's a Swashbuckle issue which we cannot fix anyway

docs/features/openapi.securitydefinitions.md
tomkerkhove: Thanks!

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
stijnmoreels: True on that we could not fix it, but at least we handle it.
Was an idea, not blocking. Maybe I'm sometimes a bit to paranoid.

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
fgheysels: You mean adding the 401 & 403 possible response types ?
I'd add them both for now and we could have a look at making it configurable when there's a real requirement or feature request to make it configurable :)

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
fgheysels: In theory, it can indeed be null and considering the defensive programming techniques it should indeed be checked.
However, I wonder if you can have a method in .NET that hasn't been declared in a type :)

fgheysels: The `GetCustomAttribute` extension method calls the `GetCustomAttributes` method behind the scenes.  I wouldn't consider it as a better way to accomplish what we're doing here.

docs/features/openapi.securitydefinitions.md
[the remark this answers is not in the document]
fgheysels: Came up with something better. :)

docs/features/openapi.securitydefinitions.md
fgheysels: True, fixed this.  Thx.

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
stijnmoreels: Well, we won't have to call `.Any` and `.OfType` three times anymore. That was what I was thinking.
Not blocking of course.

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
tomkerkhove: Yes, I agree!

docs/index.md
tomkerkhove: This is ok for now, will sweep once this one is merged.

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
tomkerkhove: Code has been changed, thanks @fgheysels 	!

src/Arcus.WebApi.OpenApi.Extensions/OAuthAuthorizeOperationFilter.cs
tomkerkhove: I think we're good to go here, resolving!

PR #54, opened 2019-07-31
FEAT: add 'ISecretProvider' with '.AddAzureKeyVault' configuration builder

### LINKED ISSUE #53: Provide integration with IConfigurationBuilder.AddAzureKeyVault
[IConfigurationBuilder.AddAzureKeyVault](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.azurekeyvaultconfigurationextensions.addazurekeyvault?view=aspnetcore-2.2#Microsoft_Extensions_Configuration_AzureKeyVaultConfigurationExtensions_AddAzureKeyVault_Microsoft_Extensions_Configuration_IConfigurationBuilder_System_String_Microsoft_Extensions_Configuration_AzureKeyVault_IKeyVaultSecretManager_) allows you to pass an `IKeyVaultSecretManager` which extends the default working.

Maybe it's interesting to provide the same capabilities as our client via `IKeyVaultSecretManager` so that we can benefit from the caching capabilities that we provide.

*Edit: This is why we'll provide an alternative to the current approach in .NET Core 2.2 ([blog post](https://blog.tomkerkhove.be/2019/07/29/the-danger-of-using-config-addazurekeyvault-in-net-core/))*

stijnmoreels:
Adds an extension to the `IConfigurationBuilder` so we can pass-in our own `ISecretProvider` when replacing configuration tokens with secret values.

Closes #53


src/Arcus.WebApi.Security/Configuration/ArcusConfigurationProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+        
+        /// <summary>
+        /// Attempts to find a value with the given key, returns true if one is found, false otherwise.
+        /// </summary>
+        /// <param name="key">The key to lookup.</param>
+        /// <param name="value">The value found at key if one is found.</param>
+        /// <returns>True if key has a value, false otherwise.</returns>
+        public override bool TryGet(string key, out string value)
+        {
+            Task<string> getSecretValueAsync = _secretProvider.Get(key);
```
tomkerkhove: Let's await it directly like we've discussed before:
```suggestion
            string secretValue = _secretProvider.Get(key).ConfigureAwait(false).GetAwaiter().GetResult();
```

No need to check for null here.

src/Arcus.WebApi.Unit/Security/Configuration/SecretProviderConfigurationBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,32 @@
...
+using Xunit;
+using Arcus.WebApi.Security.Configuration;
+using System.Collections.Generic;
+
+namespace Arcus.WebApi.Unit.Security.Configuration
+{
+    public class SecretProviderConfigurationBuilderExtensionsTests
+    {
+        [Fact]
+        public void AddKeyVault_Accesses_SecretProvider_For_Secret_Values_From_Configuration_Tokens()
```
tomkerkhove: Please align to the current naming convention that we have

src/Arcus.WebApi.Unit/Security/Configuration/SecretProviderConfigurationBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Microsoft.Extensions.Configuration;
+using Xunit;
+using Arcus.WebApi.Security.Configuration;
+using System.Collections.Generic;
+
+namespace Arcus.WebApi.Unit.Security.Configuration
+{
+    public class SecretProviderConfigurationBuilderExtensionsTests
```
tomkerkhove: Can we add a test where the configurationKey is not in the vault?

docs/features/config/replace-configuration-tokens-with-isecretprovider.md
```diff
@@ -0,0 +1,23 @@
+---
+title: "Replace configuration tokens with ISecretProvider"
+layout: default
+---
+
+The `Arcus.WebApi.Security` package provides a mechanism to use your own `ISecretProvider` implementation when building your configuration for your web application.
+
+### Usage
+
+When building your `IConfiguration`, you can call the extension `.AddAzureKeyVault` to pass in your version of the `ISecretProvider` instead on using a specific key vault store name of client.
```
tomkerkhove:
```suggestion
When building your `IConfiguration`, you can call the extension `.AddAzureKeyVault` to pass in your version of the `ISecretProvider` instead of on using a specific key vault store name of client.
```

tomkerkhove: > instead of on using a specific key vault store name of client

What do you mean with this?

docs/features/config/replace-configuration-tokens-with-isecretprovider.md
```diff
@@ -0,0 +1,23 @@
...
+
+```csharp
+ISecretProvider yourSecretProvider = ...
```
tomkerkhove: Thinking out loud, would it make sense to show how you can create a secret provider in case folks are new to this?

```suggestion
var vaultAuthenticator = new ManagedServiceIdentityAuthenticator();
var vaultConfiguration = new KeyVaultConfiguration(keyVaultUri);
var yourSecretProvider = new KeyVaultSecretProvider(vaultAuthenticator, vaultConfiguration);
```

tomkerkhove:
I'm pulling in @MassimoC to get the opinion of an Arcus consumer and @gverstraete as somebody who has not used Arcus before to see if it makes sense.

Context can be found in #53

Do you think the current documentation is clear enough for you?

docs/features/config/replace-configuration-tokens-with-isecretprovider.md
stijnmoreels: Yes, makes sense.

docs/features/config/replace-configuration-tokens-with-isecretprovider.md
stijnmoreels: Those are the default `.AddAzureKeyVault` overloads.

src/Arcus.WebApi.Security/Configuration/ArcusConfigurationProvider.cs
stijnmoreels: No, what we discussed before was that we should use `async/await` even in overloaded methods. This extra check makes sure that we guard against miss-implemented `ISecretProvider` types.

```csharp
public Task<string> Get(string secretKey)
{
    return null;
{
```

tomkerkhove: Then it would be ok to have an exception, but the current approach is also fine 👌

src/Arcus.WebApi.Unit/Security/Configuration/SecretProviderConfigurationBuilderExtensionsTests.cs
stijnmoreels: Yes, but not sure how to use this in this case?

docs/features/config/replace-configuration-tokens-with-isecretprovider.md
tomkerkhove: I know, but we should be a bit more descriptive about it.

For example:
> When building your `IConfiguration`, you can use the extension `.AddAzureKeyVault` to pass in your `ISecretProvider` instead of using the built-in [Azure Key Vault provider](https://docs.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-2.2#packages).

src/Arcus.WebApi.Unit/Security/Configuration/SecretProviderConfigurationBuilderExtensionsTests.cs
stijnmoreels: Ah, fixed this.

src/Arcus.WebApi.Unit/Security/Configuration/SecretProviderConfigurationBuilderExtensionsTests.cs
stijnmoreels: Yep, done.

tomkerkhove: Thanks!

docs/features/config/replace-configuration-tokens-with-isecretprovider.md
stijnmoreels: 👍

tomkerkhove: Thanks!

tomkerkhove:
Thank you @stijnmoreels 	 & @gverstraete

PR #55, opened 2019-07-31
Reshuffle NuGet packages to make it easier to consume them

tomkerkhove:
Reshuffle NuGet packages to make it easier to consume them:
- `Arcus.WebApi.All` should pull in all dependencies
- Split `Arcus.WebApi.Security` into smaller packages:
   - `Arcus.WebApi.Security.Authentication` contains all things related to auth.
- Depend on `Arcus.Security` v0.1 or above, not just 0.1

Shared access key auth which was shipped in 0.1 is still there but marked as obsolete. Once merged I'll create a new issue to track the removal of this for v1.0.

Docs will be reshuffled to no longer have tight coupling to namespaces.


src/Arcus.WebApi.Security/Authentication/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -1,13 +1,15 @@
...
 using GuardNet;
+using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Mvc;
 
 namespace Arcus.WebApi.Security.Authentication
 {
     /// <summary>
     /// Authentication filter to secure HTTP requests with shared access keys.
     /// </summary>
+    [Obsolete("Feature is moved to our 'Arcus.WebApi.Security.Authentication' NuGet package, please use Arcus.WebApi.Security.Authentication.SharedAccessKey.SharedAccessKeyAuthenticationAttribute")]
```
stijnmoreels: 👍 
I too always try to (1) describe the problem and (2) provide an alternative in obsolete functionality.
Cool!

tomkerkhove: Thank you for that!

It's the least we can do 🤷‍♂️

PR #57, opened 2019-07-31
Streamline documentation to make it easier to consume

tomkerkhove:
Streamline documentation to make it easier to consume and be less technical from the get-go.

- Provide a simple list of features
- Only list meta package on landing page
- Provide installation per feature

This might conflict with #55 but will align them when one or the other is merged.


docs/index.md
```diff
@@ -18,38 +18,20 @@ To install all Arcus.WebApi packages:
...
-```shell
-PM > Install-Package Arcus.WebApi.OpenApi.Extensions
-```
+For more granular packages we recommend reading the documentation.
 
 # Features
 
-## Arcus.WebApi.Security
-
-The `Arcus.WebApi.Security` package contains functionality to easily add security capabilities to an API.
```
stijnmoreels: Some "small" descriptions of each sub-section could be useful?
I like the structure; but maybe still add some small descriptions on each one?

tomkerkhove: What information would you like to be added to this view?

image

Not saying I don't want to change it, just trying to see what is expected!

tomkerkhove: As discussed it's good as it is today

PR #60, opened 2019-08-07
FEAT: create YAML file for CI build

stijnmoreels:
Adds the CI build as a YAML build (without the Github Package Registr y for now).

### DIFF
--- a/build/ci-build.yml
+++ b/build/ci-build.yml
@@ -1,0 +1,78 @@
+name: Build Arcus.Security CI
+
+trigger:
+  - master
+pr:
+  - master
+
+variables:
+  - group: 'Arcus - GitHub Package Registry'
+  - group: 'Build Configuration'
+
+  # Always use fixed version for .NET Core SDK
+  - name: 'DotNet.Sdk.Version'
+    value: '2.2.105'
+  - name: 'Project'
+    value: 'Arcus.WebApi'
+
+# Building the project:
+# - Download correct .NET Core SDK
+# - .NET Core build
+# - Publish build artifacts
+stages:
+  - stage: Build
+    jobs:
+      - job: Compile
+        pool:
+          vmImage: 'ubuntu-16.04'
+        steps:
+          - task: DotNetCoreInstaller@0
+            displayName: 'Import .NET Core SDK ($(DotNet.Sdk.Version))'
+            inputs:
+              version: '$(DotNet.Sdk.Version)'
+          - task: DotNetCoreCLI@2
+            displayName: 'Compile'
+            inputs:
+              projects: 'src/*.sln'
+              arguments: '--configuration $(Build.Configuration)'
+          - task: CopyFiles@2
+            displayName: 'Copy build artifacts'
+            inputs:
+              contents: '**/?(bin|obj)/**'
+              targetFolder: '$(Pipeline.Workspace)/build'
+          - task: PublishPipelineArtifact@0
+            displayName: 'Publish build artifacts'
+            inputs:
+              targetPath: '$(Pipeline.Workspace)/build'
+              artifactName: Build
+
+  # Testing unit tests of the project:
+  # - Download build artifacts
+  # - Download correct .NET Core SDK
+  # - .NET Core test
+  - stage: UnitTests
+    dependsOn: Build
+    condition: succeeded()
+    jobs:
+      - job: UnitTests
+        displayName: 'Run unit tests'
+        pool:
+          vmImage: 'ubuntu-16.04'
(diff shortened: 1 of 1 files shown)


build/ci-build.yml
```diff
@@ -0,0 +1,78 @@
+name: Build Arcus.WebApi CI
+
+trigger:
+  - master
+pr:
+  - master
```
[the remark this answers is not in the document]
tomkerkhove: This might be because of me, sorry for that!

However, I'm not sure if we should bother though since we don't use release branches, but can't hurt to build for every PR I guess!

What do you think @stijnmoreels 	?

build/ci-build.yml
```diff
@@ -0,0 +1,78 @@
...
+
+variables:
+  - group: 'Arcus - GitHub Package Registry'
+  - group: 'Build Configuration'
+
```
tomkerkhove: Maybe remove the new line to keep all variables close?

build/ci-build.yml
```diff
@@ -0,0 +1,78 @@
...
+            inputs:
+              targetPath: '$(Pipeline.Workspace)/build'
+              artifactName: Build
+
+  # Testing unit tests of the project:
+  # - Download build artifacts
+  # - Download correct .NET Core SDK
+  # - .NET Core test
+  - stage: UnitTests
+    dependsOn: Build
```
tomkerkhove: Let's add `displayName: Unit Test` on this level as well?

build/ci-build.yml
```diff
@@ -0,0 +1,78 @@
...
+            inputs:
+              contents: '**/?(bin|obj)/**'
+              targetFolder: '$(Pipeline.Workspace)/build'
+          - task: PublishPipelineArtifact@0
+            displayName: 'Publish build artifacts'
+            inputs:
+              targetPath: '$(Pipeline.Workspace)/build'
+              artifactName: Build
+
+  # Testing unit tests of the project:
```
tomkerkhove: Not sure if we need these comments though as the YAML itself serves as living documentation of our build.

What do you think @fgheysels 	?

build/ci-build.yml
```diff
@@ -0,0 +1,78 @@
...
+  - group: 'Arcus - GitHub Package Registry'
+  - group: 'Build Configuration'
+
+  # Always use fixed version for .NET Core SDK
+  - name: 'DotNet.Sdk.Version'
+    value: '2.2.105'
+  - name: 'Project'
+    value: 'Arcus.WebApi'
+
+# Building the project:
```
tomkerkhove: Not sure if we need these comments though as the YAML itself serves as living documentation of our build.

What do you think @fgheysels 	?

build/ci-build.yml
```diff
@@ -0,0 +1,78 @@
+name: Build Arcus.WebApi CI
```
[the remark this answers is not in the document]
tomkerkhove: image

build/ci-build.yml
fgheysels: For me, they can be removed; should the tasks in the pipeline change, one must also make sure that the documentation is kept in-sync, and I think that this is easily forgotton :)

tomkerkhove: That's my concern as well!

build/ci-build.yml
fgheysels: I think they should be removed :)

tomkerkhove: Can you remove them please @stijnmoreels 	? Thanks!

build/ci-build.yml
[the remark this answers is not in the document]
stijnmoreels: We indeed don't use release braches so it would only be when merging non-master branches. I think that a lot of times these branches are your own branches and therefore I would think that you're responsible for a green build.
I also think that if you such a scenario's happen on local branches and not remote ones; where there isn't any PR-mechanism, only your own disciplne.

That being said, I think it would hurt to run the build for every PR, it's just not directly my way of working where you merge local branches on a remote build system.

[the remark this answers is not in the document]
tomkerkhove: It does not perse have to be a local branch - If you have a "bigger" feature it sometimes happens that somebody forks your branch and contributes to it via a PR towards your branch to have a review process, however this will most-probably not be the case for Arcus.

However, I don't think it hurts to trigger build for every PR as we've all agreed upon so let's maybe just remove the restriction then?

tomkerkhove:
🏗 example - [link](https://dev.azure.com/codit/Arcus/_build/results?buildId=37047&view=results)

PR #79, opened 2019-10-23
FIX: use extension on publish NuGet YAML template

stijnmoreels:
:(


tomkerkhove:
I'm terrible at reviewing lately, sorry for that!

Any other fixes we need to do?

stijnmoreels:
> I'm terrible at reviewing lately, sorry for that!
> 
> Any other fixes we need to do?

I think this is the last one; since it's the last line of the file that had a problem.

PR #81, opened 2019-11-05
TEST: add tests for OAuth operation filter to add security definitions

stijnmoreels:


src/Arcus.WebApi.Unit/OpenApi/OAuthAuthorizeController.cs
```diff
@@ -0,0 +1,29 @@
...
+    [ApiController]
+    public class OAuthAuthorizeController : ControllerBase
+    {
+        public const string AuthorizedRoute = "oauth/authorize",
+                            NoneAuthorizedRoute = "oauth/none";
+
+        [HttpGet]
+        [Route(AuthorizedRoute)]
+        [Authorize]
+        public Task<IActionResult> GetAuthorized(HttpRequestMessage request)
```
tomkerkhove: Just return `IActionResult` since it's not doing anything async? 🤔 
```suggestion
        public IActionResult GetAuthorized(HttpRequestMessage request)
```

src/Arcus.WebApi.All/Arcus.WebApi.All.csproj
```diff
@@ -19,7 +19,6 @@
     <ProjectReference Include="..\Arcus.WebApi.Logging\Arcus.WebApi.Logging.csproj" />
     <ProjectReference Include="..\Arcus.WebApi.OpenApi.Extensions\Arcus.WebApi.OpenApi.Extensions.csproj" />
     <ProjectReference Include="..\Arcus.WebApi.Security.Authentication\Arcus.WebApi.Security.Authentication.csproj" />
-    <ProjectReference Include="..\Arcus.WebApi.Security.Extensions\Arcus.WebApi.Security.Extensions.csproj" />
```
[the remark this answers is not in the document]
tomkerkhove: Good point!

src/Arcus.WebApi.Unit/OpenApi/OAuthAuthorizeOperationFilterTests.cs
```diff
@@ -1,18 +1,85 @@
...
+            // Arrange
+            using (var client = _testServer.CreateClient())
+            // Act
+            using (HttpResponseMessage response = await client.GetAsync("swagger/v1/swagger.json"))
+            {
+                // Assert
+                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+                /* TODO:
+                    use 'OpenApiStreamReader' to load response content into a 'OpenApiDocument'
```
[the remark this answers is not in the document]
stijnmoreels: Hmm, in the startup we use v1 as swagger generation, but that's about the version of the API itself I guess. Have to check NuGet packages I guess.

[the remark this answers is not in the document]
stijnmoreels: v2.0 I think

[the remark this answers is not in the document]
tomkerkhove: Hm then it should be valid OpenAPI specs, no? Otherwise we have to fix that.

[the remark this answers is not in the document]
stijnmoreels: Hmm, apperently, I get also an error (on a different place but the same message) when I remove the OAuth operation filter.

I changed the tests to use the Open API reader and adds some logging for the diagnostic errors for now.

src/Arcus.WebApi.Unit/OpenApi/OAuthAuthorizeOperationFilterTests.cs
```diff
@@ -1,18 +1,111 @@
...
+                        "Cannot find OAuth GET operation in Open API spec file");
+
+                    OpenApiResponses oauthResponses = oauthOperation.Responses;
+                    Assert.DoesNotContain(oauthResponses, r => r.Key == "401");
+                    Assert.DoesNotContain(oauthResponses, r => r.Key == "403");
+                }
+            }
+        }
+
+        // TODO: should we add a test to check if we also include the security definitions when we add one of our authentication filters?
```
tomkerkhove: Why not, good suggestion!

src/Arcus.WebApi.Unit/OpenApi/OAuthAuthorizeOperationFilterTests.cs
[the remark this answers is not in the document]
tomkerkhove: Thanks!

We're good to go now, righ?

src/Arcus.WebApi.Unit/OpenApi/OAuthAuthorizeOperationFilterTests.cs
stijnmoreels: Not sure if it can be done, since the authentication operation filter checks 'OPERATIONS' and has no access to global defined filters (as far as I know).

src/Arcus.WebApi.Unit/OpenApi/OAuthAuthorizeOperationFilterTests.cs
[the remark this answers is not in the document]
stijnmoreels: Yep

src/Arcus.WebApi.Unit/OpenApi/OAuthAuthorizeOperationFilterTests.cs
tomkerkhove: Can you elaborate on that please?

stijnmoreels: The `IOperationFilter.Apply` gets the kind of HTTP `Operation` and the operation's 'context'. Which means, the called HTTP method. In this context, we don't have access to any "global" services like in an `ActionFilter` or `AuthenticationFilter`, which means we can't know with the information we get inside the OAuth Authorize Operation if the hosting application uses any kind of authorization filter (shared access, certificate, or maybe another all together ...); 

Therefore, my statement: not sure if we can implement this with the current state.

tomkerkhove: But we have access to `OpenApiDocument` don't we? So can't we get it like that? It should be in there.

stijnmoreels: Yeah; like I said: it can be tested; but maybe not implemented.

tomkerkhove: Oh now I get it. I think we don't have to enforce this and leave it up to the consumer to be honest.

Sorry I was confused. Think this can then remove the TODO and resolve the comment?

stijnmoreels: Yep, will do.

Issue #82, opened 2019-11-06
Provide support for querystring value in SharedAccessKeyAuthenticationFilter


SamVanhoutte:
__Problem description:__
Two scenarios:
1. __oAuth__: When implementing a custom API that handles the Authorization call back in an oAuth2 redirect, there is a state query string parameter being sent.  In order to validate the incoming request being authentic, it would be great if the operation to handle that Authorization callback could be tagged not just with the Header validation, but also with the QueryString validation.
1. __Event Grid web hooks__: At this moment, Event Grid can't still add headers for security when configuring a web hook as event handler.

__Proposed Solution__:
- Update the [SharedAccessKeyAuthenticationFilter](https://github.com/arcus-azure/arcus.webapi/blob/master/src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs) class to allow configuration for QueryString instead of Header validation as well. 
- Allow the configuration by adding a constructor overload in the [SharedAccessKeyAuthenticationAttribute](https://github.com/arcus-azure/arcus.webapi/blob/master/src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs) class, to indicate where to look for the secret.  (keeping the default to the header obviously)

tomkerkhove:
Is this something you'd like to pick up @stijnmoreels or want me to work on it?

@SamVanhoutte just to be clear - It's going to be one or the other or want to check both? If both, what if they are both defined?

stijnmoreels:
Yes, no problem, I can work on that.
Authorization info in the URL doesn't seem quite security recommended though.

SamVanhoutte:
@tomkerkhove - there's no hurry in this.  could be an easy thing for someone "new" to the project as well.
@tomkerkhove - I don't see scenarios where **both** header+querystring would be needed.
@stijnmoreels - agree I'd always prefer passing the information in the header, but sometimes the consuming app (like in my two scenarios mentioned initially) don't have or give this option.

tomkerkhove:
Good point @SamVanhoutte.

Let's give this a few weeks @stijnmoreels and if nobody picked it up, we can do it.

tomkerkhove:
René Bik will pick this up

PR #85, opened 2019-11-23
FEAT: Provide querystring suppport for SharedAccessKeyAuthenticationFilter 

renebik:
My first attempt...
I diverted a bit from the proposed solution as described in the issue; instead of creating an constructor overload with the ability to indicate where to look for the secret I created to separate attribute classes, i.e. one for Header and one for Querystring.
Next to that I made the SharedAccessKeyAuthenticationFilter class abstract and created to separate classes for Header and Querystring authorization respectively.

Also expanded the test class with appropriate testmethods.

I have left the documentation for now, first just wanted you to review this as I might have gone to far with this.

Relates to #82


stijnmoreels:
Thanks @renebik for your PR!

Altough I like your train of thought to split the configuration of authentication mechanisms, I've some concerns that I'll try to explain.

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/HeaderSharedAccessKeyAuthenticationFilter.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Arcus.WebApi.Security.Authentication.SharedAccessKey
+{
+    /// <summary>
+    /// Authentication filter to secure HTTP requests with shared access keys.
+    /// </summary>
+    /// <remarks>
+    ///     Please provide an <see cref="ISecretProvider"/> implementation in the configured services of the request.
+    /// </remarks>
+    public class HeaderSharedAccessKeyAuthenticationFilter : SharedAccessKeyAuthenticationFilter
```
stijnmoreels: A lot of duplication is happening because we have two types (filter & attribute) for each config option, + an abstracted type that don't has a lot of functionality to be an abstracted class.

stijnmoreels: We also have an `CertificateAuthentication` config option that has several config options (what of the certificate value to validate, and where the expected certificate value is located); but the attribute of that authentication mechanism is very straightforward so we can do the config somewhere else.

I would also suggest to have some kind of approach that would be in line what that. If you consider that we introduce yet another place to find the shared access key, we would need a lot of types to make that happen.

stijnmoreels: The `CertificateAuthentication` has a `...Validator` and `...Config` combination to make this happen. And an enum is used to indicate to location where to look.

stijnmoreels: It would decrease the complexity of the attribute, move the config to the `Startup` and make the maintainability better for us.

stijnmoreels: But let's hear what @tomkerkhove thinks.

renebik: @stijnmoreels first thanks for looking at this. I had not looked at the `CertificateAuthentication` implementation yet, so I will have a look at that. Reason I created the abstract class and the types per configuration is that I did not like to have `if..else` constructs to determine the type of authentication to perform.
@tomkerkhove set up a Teams meeting this afternoon to discuss some changes.

renebik:
@tomkerkhove @stijnmoreels Updated as discussed during Teams meeting.

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -8,29 +8,30 @@ namespace Arcus.WebApi.Security.Authentication.SharedAccessKey
...
         /// <param name="secretName">The name of the request header which value must match the stored secret with the same name as the header.</param>
         /// <exception cref="ArgumentException">When the <paramref name="secretName"/> is <c>null</c> or blank.</exception>
-        public SharedAccessKeyAuthenticationAttribute(string secretName) : this(secretName, secretName) { }
+        public SharedAccessKeyAuthenticationAttribute(string secretName) : this(secretName, secretName, secretName) { }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SharedAccessKeyAuthenticationAttribute"/> class.
         /// </summary>
         /// <param name="headerName">The name of the request header which value must match the stored secret.</param>
+        /// <param name="queryParameterName"></param>
```
tomkerkhove: Can you provide some documentation here please?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -8,29 +8,30 @@ namespace Arcus.WebApi.Security.Authentication.SharedAccessKey
...
         /// <param name="secretName">The name of the secret that's being retrieved using the <see cref="ISecretProvider.Get"/> call.</param>
         /// <exception cref="ArgumentException">When the <paramref name="headerName"/> is <c>null</c> or blank.</exception>
         /// <exception cref="ArgumentException">When the <paramref name="secretName"/> is <c>null</c> or blank.</exception>
-        public SharedAccessKeyAuthenticationAttribute(string headerName, string secretName) : base(typeof(SharedAccessKeyAuthenticationFilter))
+        public SharedAccessKeyAuthenticationAttribute(string headerName, string queryParameterName, string secretName) : base(typeof(SharedAccessKeyAuthenticationFilter))
         {
-            Guard.NotNullOrWhitespace(headerName, nameof(headerName), "Header name cannot be blank");
+            Guard.For<ArgumentException>(() => String.IsNullOrWhiteSpace(headerName) && String.IsNullOrWhiteSpace(queryParameterName), "Either Header name or Queryparameter name must be supplied.");
```
tomkerkhove: Suggestion to fix weird casing:
```suggestion
            Guard.For<ArgumentException>(() => String.IsNullOrWhiteSpace(headerName) && String.IsNullOrWhiteSpace(queryParameterName), "Either header name or query parameter name must be supplied.");
```

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -8,29 +8,30 @@ namespace Arcus.WebApi.Security.Authentication.SharedAccessKey
...
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
     public class SharedAccessKeyAuthenticationAttribute : TypeFilterAttribute
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="SharedAccessKeyAuthenticationAttribute"/> class.
         /// </summary>
         /// <param name="secretName">The name of the request header which value must match the stored secret with the same name as the header.</param>
         /// <exception cref="ArgumentException">When the <paramref name="secretName"/> is <c>null</c> or blank.</exception>
-        public SharedAccessKeyAuthenticationAttribute(string secretName) : this(secretName, secretName) { }
+        public SharedAccessKeyAuthenticationAttribute(string secretName) : this(secretName, secretName, secretName) { }
```
tomkerkhove: We used to support this here @stijnmoreels but maybe it's better to get rid of this, or did we have concrete demand for this?

We could also provide following overloads instead of this one:
```csharp
public SharedAccessKeyAuthenticationAttribute(string headerName) : this(headerName, null, headerName) { }
public SharedAccessKeyAuthenticationAttribute(string queryParameterName, string secretName) : this(null, queryParameterName, secretName) { }
```

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
```diff
@@ -20,21 +20,23 @@ namespace Arcus.WebApi.Security.Authentication.SharedAccessKey
...
         /// <param name="headerName">The name of the request header which value must match the stored secret.</param>
+        /// <param name="queryParameterName"></param>
         /// <param name="secretName">The name of the secret that's being retrieved using the <see cref="ISecretProvider.Get"/> call.</param>
         /// <exception cref="ArgumentException">When the <paramref name="headerName"/> is <c>null</c> or blank.</exception>
         /// <exception cref="ArgumentException">When the <paramref name="secretName"/> is <c>null</c> or blank.</exception>
-        public SharedAccessKeyAuthenticationFilter(string headerName, string secretName)
+        public SharedAccessKeyAuthenticationFilter(string headerName, string queryParameterName, string secretName)
         {
-            Guard.NotNullOrWhitespace(headerName, nameof(headerName), "Header name cannot be blank");
+            Guard.For<ArgumentException>(() => String.IsNullOrWhiteSpace(headerName) && String.IsNullOrWhiteSpace(queryParameterName), "Either Header name or Queryparameter name must be supplied.");
```
tomkerkhove: Suggestion to fix weird casing
```suggestion
            Guard.For<ArgumentException>(() => String.IsNullOrWhiteSpace(headerName) && String.IsNullOrWhiteSpace(queryParameterName), "Either header name or query parameter name must be supplied.");
```

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
```diff
@@ -20,21 +20,23 @@ namespace Arcus.WebApi.Security.Authentication.SharedAccessKey
...
     public class SharedAccessKeyAuthenticationFilter : IAsyncAuthorizationFilter
     {
-        private readonly string _headerName, _secretName;
+        private readonly string _headerName, _queryParameterName, _secretName;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SharedAccessKeyAuthenticationFilter"/> class.
         /// </summary>
         /// <param name="headerName">The name of the request header which value must match the stored secret.</param>
+        /// <param name="queryParameterName"></param>
```
tomkerkhove: Can you provide documentation here as well please?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
```diff
@@ -53,27 +55,20 @@ public async Task OnAuthorizationAsync(AuthorizationFilterContext context)
...
-                if (foundSecret == null)
+                if (requestSecretHeaders.Any(headerValue => !String.Equals(headerValue, foundSecret)))
                 {
-                    throw new SecretNotFoundException(_secretName);
+                    context.Result = new UnauthorizedResult();
                 }
-
-                if (requestSecretHeaders.Any(headerValue => !String.Equals(headerValue, foundSecret)))
+            }
+            else if (context.HttpContext.Request.Query.ContainsKey(_queryParameterName))
```
tomkerkhove: I would suggest not to use an `else if`, if the API consumer specifies both the header *and* query parameter, they both have to match.

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -14,38 +14,41 @@ public class SharedAccessKeyAuthenticationAttributeTests : IDisposable
...
-        [InlineData("not empty or whitespace", null)]
-        [InlineData("not empty or whitespace", "")]
-        [InlineData("not empty or whitespace", " ")]
+        [InlineData(null, null, "not empty or whitespace")]
+        [InlineData("", "", "not empty or whitespace")]
+        [InlineData(" ", " ", "not empty or whitespace")]
+        [InlineData("not empty or whitespace", "not empty or whitespace", null)]
+        [InlineData("not empty or whitespace", "not empty or whitespace", "")]
+        [InlineData("not empty or whitespace", "not empty or whitespace", " ")]
         public void SharedAccessKeyAttribute_WithNotPresentHeaderNameAndOrSecretName_ShouldFailWithArgumentException(
```
tomkerkhove: Test name needs to be aligned
```suggestion
        public void SharedAccessKeyAttribute_WithNotPresentHeaderNameQueryParameterNameAndOrSecretName_ShouldFailWithArgumentException(
```

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -14,38 +14,41 @@ public class SharedAccessKeyAuthenticationAttributeTests : IDisposable
...
-        [InlineData("not empty or whitespace", null)]
-        [InlineData("not empty or whitespace", "")]
-        [InlineData("not empty or whitespace", " ")]
+        [InlineData(null, null, "not empty or whitespace")]
+        [InlineData("", "", "not empty or whitespace")]
+        [InlineData(" ", " ", "not empty or whitespace")]
+        [InlineData("not empty or whitespace", "not empty or whitespace", null)]
+        [InlineData("not empty or whitespace", "not empty or whitespace", "")]
+        [InlineData("not empty or whitespace", "not empty or whitespace", " ")]
         public void SharedAccessKeyFilter_WithNotPresentHeaderNameAndOrSecretName_ShouldFailWithArgumentException(
```
tomkerkhove:
```suggestion
        public void SharedAccessKeyFilter_WithNotPresentHeaderNameQueryParameterNameAndOrSecretName_ShouldFailWithArgumentException(
```

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -152,6 +215,15 @@ private async Task<HttpResponseMessage> SendAuthorizedHttpRequestWithHeader(stri
                 return await client.SendAsync(request);
             }
         }
+        private async Task<HttpResponseMessage> SendAuthorizedHttpRequestWithQuerystring(string parameterName, string parameterValue)
```
tomkerkhove: Contains some duplication from `SendAuthorizedHttpRequestWithHeader`, we should try to refactor to reduce it by creating `SendAuthorizedHttpRequest` (or so) that creates the HttpClient and sends the provided HttpRequestMessages.

That would allow `SendAuthorizedHttpRequestWithQuerystring` to create an HttpRequestMessage with query parameter and `SendAuthorizedHttpRequestWithHeader` to add the HTTP headers

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -127,14 +130,74 @@ public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized
...
+            // Act
+            using (HttpResponseMessage response = await SendAuthorizedHttpRequestWithQuerystring(ParameterName, $"something else then {nameof(secretValue)}"))
             {
                 // Assert
                 Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
             }
         }
 
+        [Fact]
+        public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized_WhenQuerystringParameterIsNotPresent()
```
tomkerkhove:
```suggestion
        public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized_WhenQueryStringParameterIsNotPresent()
```

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -127,14 +130,74 @@ public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized
...
+            // Act
+            using (HttpResponseMessage response = await SendAuthorizedHttpRequestWithQuerystring(ParameterName, secretValue))
+            {
+                // Assert
+                Assert.NotEqual(HttpStatusCode.Unauthorized, response.StatusCode);
+            }
+        }
+
+        [Fact]
+        public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized_WhenQuerystringValueNotMatchesSecretValue()
```
tomkerkhove:
```suggestion
        public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized_WhenQueryStringValueNotMatchesSecretValue()
```

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -127,14 +130,74 @@ public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized
...
+            // Act
+            using (HttpResponseMessage response = await SendAuthorizedHttpRequestWithQuerystring(ParameterName, secretValue))
+            {
+                // Assert
+                Assert.NotEqual(HttpStatusCode.Unauthorized, response.StatusCode);
+            }
+        }
+
+        [Fact]
+        public async Task AuthorizedRoute_WithSharedAccessKey_RegisteredWithCachedSecretProvider_ShouldNotFailWithUnauthorized_WhenQuerystringValueMatchesSecretValue()
```
tomkerkhove:
```suggestion
        public async Task AuthorizedRoute_WithSharedAccessKey_RegisteredWithCachedSecretProvider_ShouldNotFailWithUnauthorized_WhenQueryStringValueMatchesSecretValue()
```

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
tomkerkhove: Or just provide capability to add headers and query parameters to one method that handles everything.

This will also be required for a new test that I'd like to add

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -127,14 +130,74 @@ public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized
...
+            using (HttpResponseMessage response =
+                await SendAuthorizedHttpRequestWithHeader(HeaderName, new[] { secretValue, $"second header value other then {nameof(secretValue)}" }))
+            {
+                // Assert
+                Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            }
+        }
+
+        [Fact]
+        public async Task AuthorizedRoute_WithSharedAccessKey__RegisteredWithSecretProvider_ShouldNotFailWithUnauthorized_WhenQuerystringValueMatchesSecretValue()
```
tomkerkhove:
```suggestion
        public async Task AuthorizedRoute_WithSharedAccessKey__RegisteredWithSecretProvider_ShouldNotFailWithUnauthorized_WhenQueryStringValueMatchesSecretValue()
```

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -127,14 +130,74 @@ public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized
...
+            _testServer.AddService<ISecretProvider>(new InMemorySecretProvider((SecretName, secretValue)));
+
+            // Act
+            using (HttpResponseMessage response = await SendAuthorizedHttpRequestWithQuerystring($"something-else-then-{nameof(ParameterName)}", "some parameter value"))
+            {
+                // Assert
+                Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            }
+        }
+
```
tomkerkhove: I would like to  suggest the following tests:
1. Request should be unauthorized if I send both the header and query parameter, but header has wrong value
2. Request should be unauthorized if I send both the header and query parameter, but query parameter has wrong value
3. Request should be authorized if I send both the header and query parameter with a valid value

tomkerkhove: Can you come up with more of them @stijnmoreels?

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationController.cs
```diff
@@ -10,8 +10,8 @@ public class SharedAccessKeyAuthenticationController : ControllerBase
     {
         [HttpGet]
         [Route("authz/shared-access-key")]
-        [SharedAccessKeyAuthentication(headerName: "x-shared-access-key", secretName: "custom-access-key-name")]
-        public Task<IActionResult> TestHardCodedConfiguredSharedAccessKeyHeaderName(HttpRequestMessage message)
+        [SharedAccessKeyAuthentication(headerName: "x-shared-access-key", queryParameterName: "api-key", secretName: "custom-access-key-name")]
+        public Task<IActionResult> TestHardCodedConfiguredSharedAccessKey(HttpRequestMessage message)
```
tomkerkhove: I might be exaggerating but I would target this endpoint and an isolated one per auth mechanism approach.

With that, I mean that every query parameter-based tests should call this one, but also a new one which only has queryParameterName configured and not `headerName` to ensure that both scenarios are covered, if I'm making myself clear?

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
stijnmoreels: | Request header | Request query | Authentication | Response |
| ----------------- | ---------------- | ---------------- | ----------- |
| invalid               | invalid              | header & query| Unauthorized |
| invalid               | valid                 | header & query| Unauthorized |   
| valid                 | invalid               | header & query| Unauthorized |
| valid                 | valid                  | header & query| Authorized |
| invalid               | invalid              | header | Unauthorized |
| invalid               | valid                 | header | Unauthorized |
| valid                 | invalid               | header | Authorized |
| valid                 | valid                  | header | Authorized|
| invalid               | invalid              | query   | Unauthorized |
| invalid               | valid                 | query   | Authorized |
| valid                  | invalid              | query   | Unauthorzed |
| valid                  | valid                 | query   | Authorized |

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: Well, it can happen that the name of the secret is not the same as the name of the header/queryParam. So, I guess we still have to support it. It seems unreasonable to force them to be the same.

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
renebik: Will do

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
renebik: So what is it your proposing here?
Add 2 new overloads as @tomkerkhove suggested or..?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: Will do

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: I thought the idea was to have one OR the other, not both together which also seems a bit odd to me. Could this actually be a requirement?

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
renebik: will look at this

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
renebik: Will add test for these cases

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationController.cs
renebik: understood, will look into this, but it also relates to a previous comment regarding both header and querystring parameter are present at same time

docs/features/security/auth/shared-access-key.md
```diff
@@ -8,7 +8,7 @@ layout: default
 The `Arcus.WebApi.Security` package provides a mechanism that uses shared access keys to grant access to a web application.
 This authentication process consists of two parts:
 
-1. Find the configured HTTP request header that contains the shared access key
+1. Find the configured HTTP request header or the configured Query parameter that contains the shared access key
```
stijnmoreels: I think we should explain more how this works with examples. "Either header or query" accompanied with filter/attribute pair for each. Now the example shows only the combination where everything is filled out.

We should also mention that when query and/or header is filled out, and we have configured header and/or query, how the application will respond.

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: @tomkerkhove , any suggestions.

tomkerkhove: > Well, it can happen that the name of the secret is not the same as the name of the header/queryParam. So, I guess we still have to support it. It seems unreasonable to force them to be the same.

It's not really clear to me what your point is here @stijnmoreels? With just one value you're assuming it's the same for query parameter, header & secret or?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
tomkerkhove: The API can configure to support both header & query parameter authentication, but it's still up to the client to choose which approach they want to use.

If you ask me, the following rules apply:
- API consumer specifies header without query parameter, header must match
- API consumer specifies query parameter without header, query parameter must match
- API consumer specifies query parameter & header, query parameter & header must match
- API consumer specifies nothing, instant revoke

_(All scenarios should also be covered with tests)_

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationController.cs
tomkerkhove: See other conversation

docs/features/security/auth/shared-access-key.md
tomkerkhove: I'm skipping docs for now until feature is done and will then do a small sweep

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
renebik: > Request header	Request query	Authentication	Response
> invalid	invalid	header & query	Unauthorized
> invalid	valid	header & query	Unauthorized
> valid	invalid	header & query	Unauthorized
> valid	valid	header & query	Authorized
> invalid	invalid	header	Unauthorized
> invalid	valid	header	Unauthorized
> valid	invalid	header	Authorized
> valid	valid	header	Authorized
> invalid	invalid	query	Unauthorized
> invalid	valid	query	Authorized
> valid	invalid	query	Unauthorzed
> valid	valid	query	Authorized

@stijnmoreels these tests imply user has influence on kind of authorization performed, but this is imo not foreseen yet. Should this be added?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
renebik: @stijnmoreels @tomkerkhove I have updated the pull request, please have a look

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: @tomkerkhove This has been updated in latest PR

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
renebik: @tomkerkhove This has been updated in latest PR

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
renebik: @stijnmoreels @tomkerkhove Ignore the previous comment. Test cases have been updated

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationController.cs
renebik: @tomkerkhove This has been updated in latest PR

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
tomkerkhove: Thanks!

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
tomkerkhove: Looks good, thanks!

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
```diff
@@ -53,35 +55,55 @@ public async Task OnAuthorizationAsync(AuthorizationFilterContext context)
...
-                    throw new KeyNotFoundException(
-                        $"No configured {nameof(ICachedSecretProvider)} or {nameof(ISecretProvider)} implementation found in the request service container. "
-                        + "Please configure such an implementation (ex. in the Startup) of your application");
-                }
 
-                string foundSecret = await userDefinedSecretProvider.Get(_secretName);
-                if (foundSecret == null)
-                {
-                    throw new SecretNotFoundException(_secretName);
+                    if (requestSecretHeaders.Any(headerValue => !String.Equals(headerValue, foundSecret)))
```
tomkerkhove: Feels like this is easier to read, no? 
```suggestion
                    if (requestSecretHeaders.Any(headerValue => string.Equals(headerValue, foundSecret, StringComparison.InvariantCultureIgnoreCase)) == false)
```

We should also be case insensitive, etc. - Maybe another good candidate for a test?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
```diff
@@ -53,35 +55,55 @@ public async Task OnAuthorizationAsync(AuthorizationFilterContext context)
...
+                    {
+                        context.Result = new UnauthorizedResult();
+                    }
                 }
 
-                if (requestSecretHeaders.Any(headerValue => !String.Equals(headerValue, foundSecret)))
+                if (!String.IsNullOrWhiteSpace(_queryParameterName) && context.HttpContext.Request.Query.ContainsKey(_queryParameterName))
                 {
-                    context.Result = new UnauthorizedResult();
+                    if (context.HttpContext.Request.Query[_queryParameterName] != foundSecret)
```
tomkerkhove: Same here, how does our attribute handle different casing? Maybe good candidate for a test?

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
```diff
@@ -127,27 +134,176 @@ public async Task AuthorizedRoute_WithSharedAccessKey_ShouldFailWithUnauthorized
...
+                await SendAuthorizedHttpRequest(AuthorizedRouteHeader, headerName: HeaderName, headerValues: new[] { secretValue, $"second header value other then {nameof(secretValue)}" }))
+            {
+                // Assert
+                Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            }
+        }
+
+        #endregion HeaderValidation
+
+        #region QueryStringValidation
```
tomkerkhove: We don't really use regions, would you mind removing them please?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
```diff
@@ -53,35 +55,55 @@ public async Task OnAuthorizationAsync(AuthorizationFilterContext context)
...
+                    {
+                        context.Result = new UnauthorizedResult();
+                    }
                 }
+
             }
-            else
+        }
+
+        private async Task<string> GetAuthorizationSecret(AuthorizationFilterContext context)
```
tomkerkhove: Nitpicking - We suffix our methods with `Async`
```suggestion
        private async Task<string> GetAuthorizationSecretAsync(AuthorizationFilterContext context)
```

I've created a PR with our minimal guidelines - https://github.com/arcus-azure/arcus/pull/40

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
tomkerkhove: Now that we have 2 approaches to define things I think we should still support it and use the configured name for header-auth only based on a secret with the same name, but mark it as obsolete.

How about this approach?
```suggestion
        [Obsolete("We now support multiple ways to specify the authentication key, we recommend using our overload instead")]
        public SharedAccessKeyAuthenticationAttribute(string secretName) : this(secretName, queryParameterName: null, secretName: secretName) { }
```

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationAttributeTests.cs
renebik: will do

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: will change

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: Aren't SharedAccessKey ment to be case-sensitive?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: See other comment

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: @tomkerkhove has this been resolved now?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
renebik: +1, will change as suggested

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
tomkerkhove: You are correct, sorry about that!

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
tomkerkhove: You are correct, sorry about that!

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
```diff
@@ -53,35 +55,55 @@ public async Task OnAuthorizationAsync(AuthorizationFilterContext context)
...
-                {
-                    throw new SecretNotFoundException(_secretName);
+                    if (requestSecretHeaders.Any(headerValue => !String.Equals(headerValue, foundSecret)))
+                    {
+                        context.Result = new UnauthorizedResult();
+                    }
                 }
 
-                if (requestSecretHeaders.Any(headerValue => !String.Equals(headerValue, foundSecret)))
+                if (!String.IsNullOrWhiteSpace(_queryParameterName) && context.HttpContext.Request.Query.ContainsKey(_queryParameterName))
```
tomkerkhove: Can we add a test that ensure that different casing for the parameter still works?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
tomkerkhove: Can we add a test that ensure that different casing for the parameter still works?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
```diff
@@ -53,35 +55,55 @@ public async Task OnAuthorizationAsync(AuthorizationFilterContext context)
...
+                context.Result = new UnauthorizedResult();
+            }
+            else
             {
-                ISecretProvider userDefinedSecretProvider = 
-                    context.HttpContext.RequestServices.GetService<ICachedSecretProvider>()
-                    ?? context.HttpContext.RequestServices.GetService<ISecretProvider>();
-                
-                if (userDefinedSecretProvider == null)
+                if (!String.IsNullOrWhiteSpace(_headerName) && context.HttpContext.Request.Headers
```
tomkerkhove: Can we add a test that ensure that different casing for the parameter still works?

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
tomkerkhove: I think this reads better
```suggestion
                    if (requestSecretHeaders.Any(headerValue => string.Equals(headerValue, foundSecret)) == false)
```

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationController.cs
```diff
@@ -15,5 +15,21 @@ public Task<IActionResult> TestHardCodedConfiguredSharedAccessKey(HttpRequestMes
         {
             return Task.FromResult<IActionResult>(Ok());
         }
+
+        [HttpGet]
+        [Route("authz/shared-access-key-header")]
```
stijnmoreels: We can use the same constants you added previously for this. Thanks.

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: Added

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: changed

src/Arcus.WebApi.Unit/Security/Authentication/SharedAccessKeyAuthenticationController.cs
renebik: done

src/Arcus.WebApi.Security.Authentication/SharedAccessKey/SharedAccessKeyAuthenticationFilter.cs
renebik: Added

docs/features/security/auth/shared-access-key.md
tomkerkhove:
```suggestion
1. Find the configured HTTP request header or the configured query parameter that contains the shared access key
```

tomkerkhove: I would indeed be a bit more clear here on the scenarios where you can:
1. Use header only
2. Use query parameter only
3. Support both
4. It will block if both are specified but only one is correct, although both are specified.

stijnmoreels: Yes, agreed. My initial comment is also still valid for this. It's not so bad that we don't have a complete docs-set for this, but initial seperation header/query would be the minimum for me.

renebik:
@tomkerkhove @stijnmoreels I have updated the docs, can you have a look?

docs/features/security/auth/shared-access-key.md
```diff
@@ -74,8 +74,47 @@ public class MyApiController : ControllerBase
     public Task<IActionResult> AuthorizedGet()
     {
         return Task.FromResult<IActionResult>(Ok());
-    }
 }
 ```
 
+## Behavior in validating shared access key parameter
```
stijnmoreels: :+1: Thanks @renebik !

tomkerkhove:
Thank you @renebik !

Issue #88, opened 2019-11-22
GitHub release task can't get initial commit


stijnmoreels:
**Describe the bug**
No correct release can be made for the Arcus webapi repo because the GitHub release task fails with an `can't find initial release` error.
All the required parameters are correctly filled out (repo name and release notes).

Example: https://dev.azure.com/codit/Arcus/_build/results?buildId=39510&view=results

**To Reproduce**
Run release of Arcus web app.

**Expected behavior**
A new GitHub release is created.

tomkerkhove:
Transfering to Arcus Web API as the issue lies here.

tomkerkhove:
This worked before so should be ok, but verified and still has admin priviledges.

PR #89, opened 2019-11-26
NuGet release should use correct repo names

tomkerkhove:
NuGet release should use correct repo names which is causing releases to fail.

Fixes https://github.com/arcus-azure/arcus.webapi/issues/88


stijnmoreels:
Thanks man!

tomkerkhove:
My pleasure! I'll clean up our 0.3.0 builds that failed as a follow-up.

Issue #67, opened 2019-09-06
Provide support for out-of-the-box correlation


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Provide support for out-of-the-box correlation on both the operation & transaction level.

Every request should allow consumers to specify a transaction id by passing the `X-Transaction-Id` header. If this is not specified, one will be generated.

Internally this transaction id and operation id can be used internally for correlating telemetry.

Every response should return the used correlation information via the following headers:
- `X-Transaction-Id` - The generated or specified transaction id
- `RequestId` - The unique id that identifies this operation.

This information should also be able to be retrieved further down the application if they wish to use it.

**Describe the solution you'd like**
Middleware should be provided that handles this for every call, and registers the information for others to use.

[Request Features](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/request-features?view=aspnetcore-2.2) are a great approach to achieve this.

The headers mentioned above should just be default headers that can be configured to use different ones if the consumers wish to do so.

**Additional context**
By registering this for other components to use it, we can use it for things like telemetry enrichments.

This should be provided via a new package, my proposal would be `Arcus.WebApi.Telemetry.Correlation`.

tomkerkhove:
I've done this before and will pick this up

tomkerkhove:
While it doesn't fully overloap with it, we should check if we should use [CorrelationId📦](https://github.com/stevejgordon/CorrelationId).

tomkerkhove:
# Design Proposal
## Configuring correlation
Registration should be very simple:
```
public void Configure(IApplicationBuilder app, IHostingEnvironment env)
{
    app.UseCorrelationId();

    if (env.IsDevelopment())
    {
        app.UseDeveloperExceptionPage();
    }

    app.UseMvc();
}
```

If they want to have more control, they can use an overload to configure the behavior:
```
public void Configure(IApplicationBuilder app, IHostingEnvironment env)
{
    app.UseCorrelationId(correlationOptions => {
        // Configure correlation
    });

    if (env.IsDevelopment())
    {
        app.UseDeveloperExceptionPage();
    }

    app.UseMvc();
}
```

### Configuration Options
It allows users to configure the following configuration options:
- `Transaction.HeaderName` (default: `X-Transaction-Id`)
- `Transaction.AllowInRequest` (default: `true`)
- `Transaction.IncludeInResponse` (default: `true`)
- `Transaction.Enabled` (default: `true`)
[...]

PR #92, opened 2019-11-27
FEAT: initial correlation setup

stijnmoreels:
Combines the core functionality of the **Correlate** NuGet library with some added functionality we want, described in #67. 
Because we work on an existing library, some functionality written here may seem strange. Most of it is to be compatible with the library, and yet provide some configuration/extensions points in the future to implement all the correlation-related features we want.

Relates to #67


tomkerkhove:
Just want to clarify that we **can** use ASP.NET Core features in the current correlation, it's just something to keep in mind over time for when we'll split the generic correlation out of Web API.

This will be done so that we can re-use the foundation for other components like messaging, while it will be up to every component to get the info and provide it to the generic correlation model.

stijnmoreels:
> Just want to clarify that we **can** use ASP.NET Core features in the current correlation, it's just something to keep in mind over time for when we'll split the generic correlation out of Web API.
> 
> This will be done so that we can re-use the foundation for other components like messaging, while it will be up to every component to get the info and provide it to the generic correlation model.

Yes, agreed.

src/Arcus.WebApi.Correlation/Correlation.cs
```diff
@@ -0,0 +1,25 @@
...
+        internal Correlation()
+        {
+            // TODO: control how the 'Request-ID' gets generated.
+            RequestId = Guid.NewGuid().ToString();
+        }
+
+        /// <summary>
+        /// Gets the unique ID information of the request.
+        /// </summary>
+        internal string RequestId { get; }
```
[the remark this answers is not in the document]
stijnmoreels: Yes

src/Arcus.WebApi.Correlation/CorrelationFactory.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using Correlate;
+using GuardNet;
+
+namespace Arcus.WebApi.Correlation 
+{
+    /// <summary>
+    /// Factory to create/clean up a <see cref="Correlation"/> and associate it with a <see cref="ICorrelationContextAccessor"/>.
+    /// </summary>
+    internal class CorrelationFactory : CorrelationContextFactory
```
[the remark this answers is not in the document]
stijnmoreels: Yep

src/Arcus.WebApi.Correlation/ICorrelationAccessor.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Correlate;
+
+namespace Arcus.WebApi.Correlation 
+{
+    /// <summary>
+    /// Provides access to the <see cref="Correlation"/>.
+    /// </summary>
+    public interface ICorrelationAccessor : ICorrelationContextAccessor
```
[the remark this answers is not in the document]
stijnmoreels: The `Correlation` is for internal use because it has a setter (from Correlate); this one is for accessing in the rest of the application.

src/Arcus.WebApi.Unit/Correlation/CorrelationTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                    
+                    string actual = GetResponseHeader(response, RequestId);
+                    Assert.NotEqual(expected, actual);
+                }
+            }
+        }
+
+        private static string GetResponseHeader(HttpResponseMessage response, string headerName)
+        {
+            (string _, IEnumerable<string> values) = Assert.Single(response.Headers, header => header.Key == headerName);
```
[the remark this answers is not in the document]
stijnmoreels: Ohh 😭

src/Arcus.WebApi.Correlation/Correlation.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using Correlate;
+
+namespace Arcus.WebApi.Correlation
+{
+    /// <summary>
+    /// Represents the correlation ID information on the incoming requests and outgoing responses.
+    /// </summary>
+    internal class Correlation : CorrelationContext
```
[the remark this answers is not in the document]
stijnmoreels: I agree that it's maybe abstract, but I don't see the added value of the `...Info`. Most of the time when adding these, I guess the original name was good enough, or a new one entirely should be used.

src/Arcus.WebApi.Correlation/CorrelateOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// <summary>
+        /// Gets or sets the request headers to retrieve the correlation id from.
+        /// </summary>
+        /// <remarks>
+        ///     The first matching header will be used.
+        /// </remarks>
+        public string[] RequestHeaders { get; set; } = { CorrelationHttpHeaders.CorrelationId, CorrelationHttpHeaders.RequestId };
+
+        /// <summary>
+        /// Gets or sets whether to include the correlation id in the response.
```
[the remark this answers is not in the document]
stijnmoreels: Both ID's

src/Arcus.WebApi.Correlation/CorrelateOptions.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Correlate.Http;
+
+namespace Arcus.WebApi.Correlation 
+{
+    /// <summary>
+    /// Options for handling correlation id on incoming requests.
+    /// </summary>
+    public class CorrelateOptions
```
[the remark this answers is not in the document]
stijnmoreels: Ah, yes; and this is a good example why two classes in the same file are somtimes an added value.

src/Arcus.WebApi.Correlation/CorrelateOptions.cs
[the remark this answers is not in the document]
stijnmoreels: Will change.

src/Arcus.WebApi.Correlation/CorrelateOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the request headers to retrieve the correlation id from.
```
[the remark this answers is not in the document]
stijnmoreels: Is from Correlate, can't do.

stijnmoreels:
I removed the `Correlate` NuGet dependency, and implemented it myself. It resulted in even less files. 💃

src/Arcus.WebApi.Correlation/CorrelationInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Represents the correlation information on the current HTTP request, accessible throughout the application.
+    /// </summary>
+    public class CorrelationInfo
+    {
+        private readonly CorrelationFeature _correlationFeature;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CorrelationInfo"/> class.
+        /// </summary>
+        public CorrelationInfo(IHttpContextAccessor contextAccessor)
```
tomkerkhove: Since we'll use this class for messaging as well, is it possible that we populate this information via the ctor instead of `IHttpContextAccessor`? Or can we introduce a `HttpCorrelationInfo` that derives from `CorrelationInfo`?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
```diff
@@ -0,0 +1,131 @@
...
+            var correlation = new CorrelationFeature(operationId, transactionId);
+            httpContext.Features.Set(correlation);
+
+            AddCorrelationResponseHeaders(httpContext, operationId, transactionId);
+
+            // TODO: on exception, should we enrich the exception.Data with the correlation info?
+            await _next(httpContext);
+        }
+
+        private (string operationId, string transactionId) DetermineCorrelationIds(HttpContext httpContext, StringValues transactionIds)
```
tomkerkhove: THinking out load - Multiple return objects and do 2 seperate things, maybe just split them into two methods?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                transactionId = transactionIds.ToString();
+            }
+
+            return (operationId, transactionId);
+        }
+
+        private void AddCorrelationResponseHeaders(HttpContext httpContext, string operationId, string transactionId)
+        {
+            if (_options.Operation.IncludeInResponse)
```
tomkerkhove: Can't we consolidate this as following?

```csharp
httpContext.Response.OnStarting(() =>
{
    if (_options.Operation.IncludeInResponse)
    {
        TryAddResponseHeader(httpContext, _options.Operation.HeaderName, operationId);
    }
    if (_options.Transaction.IncludeInResponse)
    {
        TryAddResponseHeader(httpContext, _options.Transaction.HeaderName, transactionId);
    }

    return Task.CompletedTask;
});
```

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
```diff
@@ -0,0 +1,131 @@
...
+                }
+
+                _logger.LogTrace("Correlation request header '{HeaderName}' found with transaction ID '{TransactionId}'", _options.Transaction.HeaderName, transactionIds);
+            }
+
+            (string operationId, string transactionId) = DetermineCorrelationIds(httpContext, transactionIds);
+            var correlation = new CorrelationFeature(operationId, transactionId);
+            httpContext.Features.Set(correlation);
+
+            AddCorrelationResponseHeaders(httpContext, operationId, transactionId);
```
tomkerkhove: Maybe we should do this after `await _next(httpContext);` in case they start returning a response or so?

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <summary>
+    /// Adds operation and transaction correlation to the application.
+    /// </summary>
+    // ReSharper disable once InconsistentNaming
+    public static class IServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Adds the correlation related services to the dependency injection system of the application.
+        /// </summary>
+        public static IServiceCollection AddCorrelation(this IServiceCollection services, Action<CorrelationOptions> configureOptions)
```
tomkerkhove: Configuring the correlation should be optional, no?
```suggestion
        public static IServiceCollection AddCorrelation(this IServiceCollection services, Action<CorrelationOptions> configureCorrelationOptions = null)
```

stijnmoreels: Yes, that's why there's an overload for this.

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Well, shouldn't that response then have the correlation headers?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Yeah was also thinking, but I don't know the lifecycle of the options instance, that's why I moved it outside the lambda so we don't get capture-variable problems.

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Yeah, thanks.

src/Arcus.WebApi.Correlation/CorrelationInfo.cs
stijnmoreels: I renamed the `CorrelationFeature` public and renamed it to `CorrelationInfo` + I renamed the `CorrelationInfo` to `HttpCorrelationInfo`. That way, we can reuse the `CorrelationInfo` class in messaging.

src/Arcus.WebApi.Correlation/ICorrelationAccessor.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        /// <summary>
+        /// Gets the transactional ID of the request.
+        /// </summary>
+        string CorrelationId { get; }
```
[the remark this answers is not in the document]
stijnmoreels: Done.

src/Arcus.WebApi.Correlation/CorrelateOptions.cs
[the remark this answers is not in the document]
stijnmoreels: Done.

src/Arcus.WebApi.Correlation/CorrelateOptions.cs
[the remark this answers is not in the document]
stijnmoreels: Removed 'Correlate' dependency, not related anymore.

src/Arcus.WebApi.Correlation/CorrelateOptions.cs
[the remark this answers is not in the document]
stijnmoreels: Removed 'Correlate' dependency, not related anymore.

src/Arcus.WebApi.Correlation/CorrelateOptions.cs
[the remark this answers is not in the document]
stijnmoreels: Yep, done.

src/Arcus.WebApi.Correlation/Correlation.cs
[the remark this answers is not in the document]
stijnmoreels: Added anyway.

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace Arcus.WebApi.Correlation 
+{
+    /// <summary>
+    /// Adds the correlation related services to the <see cref="IServiceCollection"/> of the application.
+    /// </summary>
+    // ReSharper disable once InconsistentNaming
+    public static class IServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Adds the correlation related services to the dependency injection system of the application.
```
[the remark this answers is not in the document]
stijnmoreels: Done.

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            AddCorrelation(services);
+
+            services.Configure(configureOptions);
+
+            return services;
+        }
+
+        /// <summary>
+        /// Adds the correlation related services to the dependency injection system of the application.
```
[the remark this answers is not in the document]
stijnmoreels: Done.

src/Arcus.WebApi.Unit/Correlation/CorrelationTests.cs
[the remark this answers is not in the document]
stijnmoreels: Done.

src/Arcus.WebApi.Correlation/CorrelationAccessor.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Correlate;
+using GuardNet;
+
+namespace Arcus.WebApi.Correlation 
+{
+    /// <summary>
+    /// Represents a singleton <see cref="ICorrelationAccessor"/> implementation to store the correlation ID information.
+    /// </summary>
+    public class CorrelationAccessor : ICorrelationAccessor
```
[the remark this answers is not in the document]
stijnmoreels: Removed 'Correlate' as dependency, not related anymore.

src/Arcus.WebApi.Correlation/ICorrelationAccessor.cs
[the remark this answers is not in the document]
stijnmoreels: Removed 'Correlate' as dependecy, not related anymore.

src/Arcus.WebApi.Correlation/IApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Arcus.WebApi.Correlation 
+{
+    /// <summary>
+    /// Adds the <see cref="CorrelationMiddleware"/> to the application pipeline.
+    /// </summary>
+    // ReSharper disable once InconsistentNaming
+    public static class IApplicationBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the <see cref="CorrelationMiddleware"/> to the application pipeline.
```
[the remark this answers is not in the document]
stijnmoreels: Done.

src/Arcus.WebApi.Correlation/CorrelationFactory.cs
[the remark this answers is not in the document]
stijnmoreels: Removed 'Correlate' as dependency, not related anymore.

src/Arcus.WebApi.Correlation/Correlation.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Correlation"/> class.
+        /// </summary>
+        internal Correlation()
+        {
+            // TODO: control how the 'Request-ID' gets generated.
+            RequestId = Guid.NewGuid().ToString();
```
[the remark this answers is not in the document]
stijnmoreels: Done now in middleware.

src/Arcus.WebApi.Correlation/Correlation.cs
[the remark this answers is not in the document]
stijnmoreels: Done.

src/Arcus.WebApi.Correlation/CorrelationInfo.cs
tomkerkhove: Perfect, thanks!

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
tomkerkhove: Good for me!

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
tomkerkhove: That's a good point, but how it is now, do we actually return the operation id or not? Because we are calling `OnStarting` twice which feels a bit odd

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    TryAddResponseHeader(httpContext, _options.Transaction.HeaderName, transactionId);
+                    return Task.CompletedTask;
+                });
+            }
+        }
+
+        private void TryAddResponseHeader(HttpContext httpContext, string headerName, string headerValue)
+        {
+            if (!httpContext.Response.Headers.ContainsKey(headerName))
```
tomkerkhove: Should we check of just overwrite? I'd say the latter, no?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <returns>A <see cref="T:System.Threading.Tasks.Task" /> that represents the execution of this middleware.</returns>
+        public async Task Invoke(HttpContext httpContext)
+        {
+            Guard.NotNull(httpContext, nameof(httpContext));
+            Guard.For<ArgumentException>(() => httpContext.Response is null, "Requires a 'Response'");
+            Guard.For<ArgumentException>(() => httpContext.Response.Headers is null, "Requires a 'Response' object with headers");
+
+            if (httpContext.Request.Headers.TryGetValue(_options.Transaction.HeaderName, out StringValues transactionIds))
+            {
+                if (!_options.Transaction.AllowInRequest)
```
tomkerkhove: Thinking out loud:

Do we want to check if it's allowed or just if we need to take it into account?

In other words, do we block the header if specified or do we just ignore it?

tomkerkhove:
Added a few comments but looks nice already! Almost there!

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Yeah; both are added. The `OnStarting` _Adds_ a delegate.

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: If we ignore it; I wouldn't call it `Allow...` since that would mean it _would_ block. No?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Yeah, makes sense. I'll change that.

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
tomkerkhove: Good point!

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
tomkerkhove: Yes, my question is more from a functional standpoint - What would you like?

But blocking is ok for me.

tomkerkhove:
I'm fine with either to be honest! Just to be clear, which breaking change are you talking about @fgheysels?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Yeah, I like it when we shortcut it if we 'don't allow' such headers.

tomkerkhove: Sounds good to me

stijnmoreels:
> * I don't think all .NET languages support optional parameters.  When you have a look at the .NET framework / .NET Core libs, I don't think you'll encounter the use of optional parameters.

F# supports this, internally mapped to an optional type. But I see your point since this will be a 'netstandard' library.

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+
+            string operationId = DetermineOperationId(httpContext);
+            string transactionId = DetermineTransactionId(httpContext, transactionIds);
+            var correlation = new CorrelationInfo(operationId, transactionId);
+            httpContext.Features.Set(correlation);
+
+            AddCorrelationResponseHeaders(httpContext, operationId, transactionId);
+
+            // TODO: on exception, should we enrich the exception.Data with the correlation info?
```
tomkerkhove: I'd say no to be honest as the headers are already in the response itself?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
```diff
@@ -0,0 +1,130 @@
...
+            await _next(httpContext);
+        }
+
+        private static string DetermineOperationId(HttpContext httpContext)
+        {
+            // TODO: make ID generation configurable for the consumer.
```
tomkerkhove: Let's add an issue for this and remove the todo?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
```diff
@@ -0,0 +1,130 @@
...
+            return httpContext.TraceIdentifier ?? Guid.NewGuid().ToString();
+        }
+
+        private string DetermineTransactionId(HttpContext httpContext, StringValues transactionIds)
+        {
+            // TODO: make ID generation configurable for the consumer.
```
tomkerkhove: Let's add an issue for this and remove the todo?

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Ok, good.

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Done #102

src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs
stijnmoreels: Done #103

tomkerkhove:
> I see the benefit of using optional parameters over method overloading but there are a few things that must be taken into consideration:
> 
> * this change is a breaking change; this version of the library will not be binary compatible with previous versions.  This means that consumers will need to recompile if they use this version.   This is imho not a problem, given the fact that most consumers will use this as a nuget package and will recompile anyway.
> * I don't think all .NET languages support optional parameters.  When you have a look at the .NET framework / .NET Core libs, I don't think you'll encounter the use of optional parameters.

What is your preference here @stijnmoreels? While we wait for @fgheysels we can also move to overloads if you prefer that as well.

src/Arcus.WebApi.Correlation/CorrelationOptionsOperation.cs
```diff
@@ -0,0 +1,33 @@
+﻿using GuardNet;
+
+namespace Arcus.WebApi.Correlation
+{
+    /// <summary>
+    /// Correlation options specific for the operation ID.
+    /// </summary>
+    public class CorrelationOptionsOperation
+    {
+        private string _headerName = "X-Operation-ID";
```
tomkerkhove: Should be `RequestId`
```suggestion
        private string _headerName = "RequestId";
```

stijnmoreels: Done.

stijnmoreels:
> > I see the benefit of using optional parameters over method overloading but there are a few things that must be taken into consideration:
> > 
> > * this change is a breaking change; this version of the library will not be binary compatible with previous versions.  This means that consumers will need to recompile if they use this version.   This is imho not a problem, given the fact that most consumers will use this as a nuget package and will recompile anyway.
> > * I don't think all .NET languages support optional parameters.  When you have a look at the .NET framework / .NET Core libs, I don't think you'll encounter the use of optional parameters.
> 
> What is your preference here @stijnmoreels? While we wait for @fgheysels we can also move to overloads if you prefer that as well.

Don't know exactly. I'm fine with either; but I think we should stick with an approach, and since we've already used optional arguments, I'm not sure what to do next.

stijnmoreels:
> LGTM!
> 
> Docs are not there yet but it's ok for me to do it in a dedicated PR.
> Holding off with merging until we have a reply from Frederik.

I'll add the docs while we wait on @fgheysels .

docs/index.md
```diff
@@ -26,6 +26,8 @@ For more granular packages we recommend reading the documentation.
     - Authentication
         - [Shared access key authentication](features/security/auth/shared-access-key)
         - [Certificate authentication](features/security/auth/certificate)
+- **Correlation**
+    - [Correlate between request/responses](feature/correlation)
```
tomkerkhove:
```suggestion
    - [Correlate between request/responses](features/correlation)
```

Maybe we should be more precise and elaborate what we correlate?
`Provide request and/or transaction correlation ids`?

docs/features/correlation.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+Note: because the correlation is based on <span>ASP.NET</span> Core middleware, it's recommended to place it before the `.UseMvc` call.
+
+## Configuration
+
+Some extra options are available to alter the functionality of the correlation:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
```
tomkerkhove: This is great, but we should mention what the defaults are

docs/features/correlation.md
```diff
@@ -0,0 +1,85 @@
...
+## How This Works
+
+When an application is configured to use the default configuration of the correlation, each HTTP response will get an extra header called `RequestId` containing an unique identifier to distinguish between requests/responses.
+The `X-Transaction-ID` can be overriden by the request, meaning: if the HTTP request already contains a `X-Transaction-ID` header, the same header+value will be used in the HTTP response.
+
+Additional [configuration](#configuration) is available to tweak this functionality.
+
+## Installation
+
+This feature requires to install our NuGet package
```
tomkerkhove:
```suggestion
This feature requires to install our NuGet package:
```

docs/features/correlation.md
```diff
@@ -0,0 +1,85 @@
+---
+title: "Correlate between HTTP requests/responses via ASP.NET Core middleware"
+layout: default
+---
+
+# Correlation Between HTTP Requests
+
+The `Arcus.WebApi.Correlation` library provides a way to add correlation between HTTP requests. 
+This correlation happens with the use of two HTTP request/response headers: `RequestId` and `X-Transaction-ID`.
```
tomkerkhove:
```suggestion

This correlation is based the `RequestId` and `X-Transaction-ID` HTTP request/response headers, however, you can fully configure different headers in case you need to.
```

docs/features/correlation.md
```diff
@@ -0,0 +1,85 @@
...
+
+## How This Works
+
+When an application is configured to use the default configuration of the correlation, each HTTP response will get an extra header called `RequestId` containing an unique identifier to distinguish between requests/responses.
+The `X-Transaction-ID` can be overriden by the request, meaning: if the HTTP request already contains a `X-Transaction-ID` header, the same header+value will be used in the HTTP response.
```
tomkerkhove:
```suggestion

The `X-Transaction-ID` can be overriden by the request, meaning: if the HTTP request already contains a `X-Transaction-ID` header, the same header+value will be used in the HTTP response.
```

docs/index.md
stijnmoreels: 👍

docs/features/correlation.md
stijnmoreels: Yeah, hmm. all the values set here are the defaults; but I guess indeed that's not clear.
Maybe a table of some sort with type+default?

docs/features/correlation.md
stijnmoreels: Yes 🥇 , that's more clear.

docs/features/correlation.md
tomkerkhove: You can also keep the current approach and just add a `(default: true)`?

If we would switch to a table, I would remove the docs below then and use table instead.

Up to you to decide, both are ok.

stijnmoreels: Ah, yes. Thanks. That would be better.

tomkerkhove: Nitpick - Maybe move it at the end instead of beginning?
Is a bit weird to read.

stijnmoreels: No problem, will do.

fgheysels:
> I'm fine with either to be honest! Just to be clear, which breaking change are you talking about @fgheysels?

As said in my initial comment, I don't think it is much of an issue in this particular case since you'll normally recompile anyway when updating the package.
I just had a 'blast from the past' where I had an application that used a certain dll.   That dll had overload methods, and I've changed them to optional parameters.   Since I had not made any other changes, I just dropped that modified dll and thought that would be OK.   As you can imagine, it was not OK since the calling application was no longer able to find that method since its signature had changed.   (That was in an era where build & release pipelines was not something that we commonly used :) ).

tomkerkhove:
Ha, that's good to know, thanks!

In meantime we've aligned with your suggestion and will certainly keep this in mind!

PR #93, opened 2019-11-29
CHORE: update Arcus.Security to v1

### LINKED ISSUE #90: Use Arcus Security v1.x
We should move to use Arcus Security v1.x and see if we need to make any breaking changes.

stijnmoreels:
Upgrade the WebApi authentication project to the v1 of the Arcus.Security package.

Closes #90


tomkerkhove:
That's a tricky one - If we merge this, it's going to be a new major version, no 🤔

stijnmoreels:
> That's a tricky one - If we merge this, it's going to be a new major version, no 🤔

Hmm, ah, yes; maybe place this on hold for a while.

Issue #68, opened 2019-09-06
Provide a Serilog enricher for out-of-the-box correlation information


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Serilog allows you to write [enrichers](https://github.com/serilog/serilog/wiki/Enrichment) that will enrich every entry without having to worry about it.

An enricher should be introduced that does this for our out-of-the-box correlation (#67)

**Describe the solution you'd like**
Write a simple [``](https://github.com/serilog/serilog/wiki/Enrichment) that consumes the correlation information that was made available by #67.

**Additional context**
This should be isolated in a package specific for Serilog & telemetry such as `Arcus.WebApi.Telemetry.Serilog`.

tomkerkhove:
I've done this before and will pick this up

stijnmoreels:
The correlation library that we use (`Correlate`) already has some logging for every request of the correlation ID. Maybe this feature should somewhat be altered if we indeed continue to use `Correlate`?

tomkerkhove:
Let's see where we land with that first ;)

stijnmoreels:
It's not gonna be easy...
Either we don't use the recommended approach to use Serilog and get access to the `IHttpContextAccessor` to get each time the correct correlation, or we use the recommended approach to use Serilog but we don't use enricher implementation but add a function to the `EnrichDiagnosticContext` of the `app.UseSerilogRequestLogging()` call.

tomkerkhove:
I'd go for using our own interface that we've made so that it works with WebApi & Workers, but ok if it just works with workers for now.

Maybe we can create a small POC for this as well and see what it looks like?

stijnmoreels:
> I'd go for using our own interface that we've made so that it works with WebApi & Workers, but ok if it just works with workers for now.
> 
> Maybe we can create a small POC for this as well and see what it looks like?

Hmm, then we maybe move this issue to messaging? For workers?

tomkerkhove:
It covers both so we don't have a good story for that yet, let's just keep it here for now

stijnmoreels:
This is done?
If so, we could release v0.4.0?

tomkerkhove:
The question is, do we ship it and move it over later or do we do that already?
https://github.com/arcus-azure/arcus.observability/issues/18

Otherwise it's already a deprecation we could've avoid.

stijnmoreels:
Ah, we should probably move it then.

tomkerkhove:
[...]

PR #99, opened 2020-01-02
FEAT: add Serilog enrichment with correlation info

stijnmoreels:
Adds simple Serilog enrichment for the correaltion feature.

Relates to #68


src/Arcus.WebApi.sln
```diff
@@ -17,7 +17,16 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Arcus.WebApi.OpenApi.Extens
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Arcus.WebApi.Security.Authentication", "Arcus.WebApi.Security.Authentication\Arcus.WebApi.Security.Authentication.csproj", "{88EADCAB-29BF-42F1-8748-646676898903}"
 EndProject
+<<<<<<< HEAD
```
tomkerkhove: This doesn't look good :D

src/Arcus.WebApi.Telemetry.Serilog/Arcus.WebApi.Telemetry.Serilog.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <Authors>Arcus</Authors>
+    <Company>Arcus</Company>
+    <RepositoryType>Git</RepositoryType>
+    <PackageTags>Azure;WebAPI;App Services;Web App;Web;API</PackageTags>
+    <Description>Provides capabilities with serilog telemetry for Web APIs running in Azure.</Description>
```
tomkerkhove: How about this?
```suggestion
    <Description>Provides better integration with Serilog for emitting Web API telemetry running in Azure.</Description>
```

src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// </summary>
+    public class CorrelationInfoEnricher : ILogEventEnricher
+    {
+        private readonly HttpCorrelationInfo _correlationInfo;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CorrelationInfoEnricher"/> class.
+        /// </summary>
+        /// <param name="correlationInfo">The correlation information registered in the <see cref="IServiceCollection"/>.</param>
+        public CorrelationInfoEnricher(HttpCorrelationInfo correlationInfo)
```
tomkerkhove: We need to be careful here, will an instance be created every time or will it only be initialized once?

In other words, will this work correctly at runtime or always re-use correlation info?

src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// Enrich the log event.
+        /// </summary>
+        /// <param name="logEvent">The log event to enrich.</param>
+        /// <param name="propertyFactory">Factory for creating new properties to add to the event.</param>
+        public void Enrich(LogEvent logEvent, ILogEventPropertyFactory propertyFactory)
+        {
+            if (!String.IsNullOrEmpty(_correlationInfo.TransactionId))
+            {
+                LogEventProperty transactionIdProperty = 
+                    propertyFactory.CreateProperty(nameof(_correlationInfo.TransactionId), _correlationInfo.TransactionId);
```
tomkerkhove: I'd use a constant for this so that renames are not impacting this.
```suggestion
                    propertyFactory.CreateProperty(TransactionIdProperty, _correlationInfo.TransactionId);
```

src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                logEvent.AddPropertyIfAbsent(transactionIdProperty);
+            }
+
+            LogEventProperty operationIdProperty =
```
tomkerkhove: I would, also here, check if it's null or not and use a constant instead of `nameof`

src/Arcus.WebApi.sln
stijnmoreels: 😄 yeah, but isn't that the outdated version?

tomkerkhove: Nevermind me 🙄

src/Arcus.WebApi.Telemetry.Serilog/Arcus.WebApi.Telemetry.Serilog.csproj
stijnmoreels: All good.

src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs
stijnmoreels: I think it will use `"TransactionId"` when you use `nameof(_correlationInfo.TransactionId)`.
Like, when using a namespace: `nameof(System.TimeSpan)` the result is `"TimeSpan"`.
But, you're right. I'll will be better.

src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs
stijnmoreels: Yep, ok.

src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs
```diff
@@ -1,54 +1,39 @@
...
+using Microsoft.AspNetCore.Http;
+using Serilog;
 
 namespace Arcus.WebApi.Telemetry.Serilog.Correlation
 {
     /// <summary>
     /// Enriches the log events with the correlation information.
     /// </summary>
-    public class CorrelationInfoEnricher : ILogEventEnricher
+    public class CorrelationInfoEnricher : IDiagnosticEnricher
```
tomkerkhove: This is ok, but you're limiting it to HTTP requests, what about my traces?

stijnmoreels: It's probably got to be somewhere else.

tomkerkhove: Why can't we use the `ILogEventEnricher` which covers all telemetry?

stijnmoreels: The request logging has the `IDiagnosticContext` and `HttpContext` as input parameters, the othe `ILogEventEnricher` has a `LogEvent` as input parameter.

tomkerkhove: That's right but it doesn't cover everything, so we have to use `ILogEventEnricher` to support non-HTTP request telemetry as well

stijnmoreels: So how should approach this? Like implementing the interface but not use them in the request logging?

tomkerkhove: How about using `ILogEventEnricher` and use `services.GetService<CorrelationInfo>` for every trace?

I think that's the only way to get it working but it covers all scenarios, at least for now

tomkerkhove: Just FYI - `ILogEventEnricher` will be called for every single message that is being written

stijnmoreels: Yes, but than the only question is to get hold of the service provider since it's being set further up. It would require us to step away from the Serilog recommended approach of registering the logger.

tomkerkhove: That's why it's a recommendation ;)

There is no other way so we have to go there.

stijnmoreels: Let me try something...

stijnmoreels: Now it uses the Serilog `ILogEventEnricher` but uses the `HttpCorrelationInfo` because that's available in the services. The `CorrelationInfo` is added as a request feature during the receival of the request.

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using static Arcus.WebApi.Unit.Correlation.CorrelationController;
+
+namespace Arcus.WebApi.Unit.Correlation
+{
+    public class TelemetryCorrelationTests
+    {
+        private readonly TestApiServer _testServer = new TestApiServer();
+
+        [Fact]
+        public async Task SendRequest_WithSerilogCorrelationRequestLogging_ReturnsOkWithEncrichedCorrelationLogProperties()
```
tomkerkhove: Can we extend it with:
- A test to verify that it uses 2 different operation ids but same transaction id?
- A test to verify that it uses 2 different operation ids & transaction id?

src/Arcus.WebApi.Unit/Correlation/CorrelationController.cs
```diff
@@ -10,13 +15,15 @@ public class CorrelationController : ControllerBase
         public const string Route = "correlation";
 
         private readonly HttpCorrelationInfo _correlationInfo;
+        private readonly DiagnosticContext _diagnosticContext;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CorrelationController"/> class.
         /// </summary>
-        public CorrelationController(HttpCorrelationInfo correlationInfo)
+        public CorrelationController(HttpCorrelationInfo correlationInfo, DiagnosticContext diagnosticContext)
```
tomkerkhove: Can be removed I think, no?

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
stijnmoreels: Yes, thanks, that would make it indeed better.

src/Arcus.WebApi.Unit/Correlation/CorrelationController.cs
stijnmoreels: Hmm, I added this to have a way to retrieve the enriched log properties from a Controller endpoint. So, we would still require this if we want to verify the properties.

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
stijnmoreels: That was added.

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
```diff
@@ -33,7 +33,7 @@ public static IServiceCollection AddCorrelation(this IServiceCollection services
             Guard.NotNull(services, nameof(services));
 
             services.AddHttpContextAccessor();
-            services.AddSingleton<HttpCorrelationInfo>();
+            services.AddTransient<HttpCorrelationInfo>();
```
tomkerkhove: Transient or [scoped](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#scoped)?

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+using Arcus.WebApi.Correlation;
+using Arcus.WebApi.Unit.Hosting;
+using Newtonsoft.Json;
+using Serilog.Events;
+using Xunit;
+using static Arcus.WebApi.Unit.Correlation.CorrelationController;
+
+namespace Arcus.WebApi.Unit.Correlation
```
tomkerkhove: I'd suggest one more test - What if you do 2 sequential calls, do they get the right correlation info?

src/Arcus.WebApi.Unit/Correlation/CorrelationController.cs
tomkerkhove: Ok, it's testing so 🤷‍♂

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
stijnmoreels: I think transient is more safe/stable in this context, no?

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
stijnmoreels: How you mean? Like two independent calls with both diff operation/transaction ID's?

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
tomkerkhove: Not really because the correlation info will be re-created every time instead of on a per-request basis, unless I'm missing something?

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
tomkerkhove: 2 independent GETs rather than just one to verify if the correlation info is set correctly for both calls

stijnmoreels: Isn't this test: `SendRequest_WithSerilogCorrelationEnrichment_ReturnsOkWithDifferentOperationIdAndDifferentTransactionId` doing just that?

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
stijnmoreels: Hmm, the tests fails if I change it to scoped. The log properties are in that case a little behind.

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
tomkerkhove: Yes, I've missed that one sorry.

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            var content = JsonConvert.DeserializeAnonymousType(json, new { TransactionId = "", OperationId = "" });
+            Assert.False(String.IsNullOrWhiteSpace(content.TransactionId), "Accessed 'X-Transaction-ID' cannot be blank");
+            Assert.False(String.IsNullOrWhiteSpace(content.OperationId), "Accessed 'X-Operation-ID' cannot be blank");
+
+            return new CorrelationInfo(content.OperationId, content.TransactionId);
+        }
+
+        private void AssertLoggedCorrelationProperties(CorrelationInfo correlationInfo)
+        {
+            IEnumerable<KeyValuePair<string, LogEventPropertyValue>> properties =
```
tomkerkhove: This is tricky, if we do a 2nd request it will still contain the traces from the first call, so the test can be successful while it's wrong, no?

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            // Arrange
+            using (HttpClient client = _testServer.CreateClient())
+            // Act
+            using (HttpResponseMessage firstResponse = await client.GetAsync(Route))
+            {
+                // Assert
+                Assert.Equal(HttpStatusCode.OK, firstResponse.StatusCode);
+                CorrelationInfo firstCorrelationInfo = await AssertAppCorrelationInfoAsync(firstResponse);
+                AssertLoggedCorrelationProperties(firstCorrelationInfo);
```
tomkerkhove: We should pass a transaction id in the request and also verify if that matches what was found

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
stijnmoreels: Yes, that's why I called the method `Dequeue...` because it removes the log events in the in-memory sink after it's called.

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
stijnmoreels: Isn't that what happens in the test above this one?

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
tomkerkhove: Ah my bad, sorry - Looks like I'm having a bad day :/

src/Arcus.WebApi.Unit/Correlation/TelemetryCorrelationTests.cs
tomkerkhove: Ah my bad, sorry - Looks like I'm having a bad day :/

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
tomkerkhove: Hm I'll trust the tests then, but it's strange :D

src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs
tomkerkhove: Nice, thanks!

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
stijnmoreels: We could look into it deeper if you want, afterwards. Issue for this?

src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>Enrich the log event.</summary>
+        /// <param name="logEvent">The log event to enrich.</param>
+        /// <param name="propertyFactory">Factory for creating new properties to add to the event.</param>
+        public void Enrich(LogEvent logEvent, ILogEventPropertyFactory propertyFactory)
+        {
+            Guard.NotNull(logEvent, nameof(logEvent));
+            Guard.NotNull(propertyFactory, nameof(propertyFactory));
+
+            var correlationInfo = _services.GetRequiredService<HttpCorrelationInfo>();
```
tomkerkhove: Thinking out loud - Can we make this depend on `CorrelationInfo` so that we can move it to Arcus Observation eventually?

stijnmoreels: I know. The problem is that the HTTP variant is a service and the correlation is a request feature. Maybe if we could make the correlation a scoped service (for each request). If that's possible, and at least as safe as the request features.

tomkerkhove: Let's keep it as-is and see how we'll do that later on!

src/Arcus.WebApi.Correlation/IServiceCollectionExtensions.cs
tomkerkhove: I get how it works now, I think it's ok for now

PR #106, opened 2020-01-20
DOCS: add feature doc for automatically invalidate Azure Key Vault secrets

stijnmoreels:
Relates to #101, #86, #96

Issue #96, opened 2019-12-04
Provide background job to securely receive CloudEvents


tomkerkhove:
Provide background job to securely receive CloudEvents.

This allows APIs to have asynchronous jobs that allow external parties to forward events to react upon, without exposing a public endpoint.

A good example of this is #86 which will run on this behind the scenes:
![image](https://user-images.githubusercontent.com/4345663/70146461-7912a700-16a2-11ea-9361-e972d4eab154.png)

This generic background job would receive the `CloudEvent` but not run any custom code as this will be up to the implementing job to handle the event itself.

stijnmoreels:
@tomkerkhove could you internally add an event subscription on the KeyVault Azure resource, to write the `New Secret Version Created` events to a ServiceBus Queue; that way, I don't have to use my Azure subscription for that.

tomkerkhove:
Events from `arcus-security-dev-we` Azure Key Vault should emit `Microsoft.KeyVault.SecretNewVersionCreated` events to `https://arcus-webapi-dev-we-integration-tests.servicebus.windows.net/new-secret-version-events`.

`ServiceBus.Topics.KeyVaultEvents.ConnectionString` variable is added to the `Arcus Web API - Integration Testing` variable group with a connection string scoped to the topic.

PR #110, opened 2020-01-20
FEAT: move 'CloudEventBackgroundJob' to sep project

stijnmoreels:
Relates to #96 , #106

### DIFF
--- a/src/Arcus.BackgroundJobs/CloudEvent/CloudEventBackgroundJob.cs
+++ b/src/Arcus.BackgroundJobs/CloudEvent/CloudEventBackgroundJob.cs
@@ -12,12 +12,12 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
-namespace Arcus.WebApi.Jobs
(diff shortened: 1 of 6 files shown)


src/Arcus.BackgroundJobs/CloudEvent/CloudEventBackgroundJob.cs
```diff
@@ -12,12 +12,12 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
-namespace Arcus.WebApi.Jobs
+namespace Arcus.BackgroundJobs.CloudEvent
```
tomkerkhove: Do we use this, `Arcus.BackgroundJobs` or `Arcus.BackgroundJobs.Core`?

stijnmoreels: IDK, I used this specific sub-namespace because it would maybe make more sense when other types of jobs are added. Each in a sub-topic-namespace?

tomkerkhove: Sounds good to me!

PR #117, opened 2020-02-25
FEAT : built-in request filter exception handling

### LINKED ISSUE #119: Take into account the built-in `BadHttpRequestException.StatusCode` when handling exceptions
When a built-in restriction is violated, an `BadHttpRequestException` is thrown. We could benefit from its `.StatusCode` property to respond to the client a more describitive error.

stijnmoreels:
When using for example the built-in [max request body size option](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserverlimits.maxrequestbodysize?view=aspnetcore-3.1), this will result in an `BadHttpRequestException` which internally uses a HTTP status code to indicate which sub-type of a bad request is happening.

Having our own exception handling to know about this behavior, we can interact with it and re-use that HTTP status code to provide a more descripitive response instead of always the 500 code.

Again, proposal. We require an extra NuGet package for this, so not sure if it's worth it.
The extra test works for .NET Core 2.2 because then the exception is publicly available, in 3.+ this is internal/private. But since the exception handling logic is written in netstandard, there's no problem doing testing like this I suppose.

Closes #119


stijnmoreels:
Speaking of the max request size, should we provide one in the web API template so people are automatically thinking about it?

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
```diff
@@ -62,20 +64,24 @@ public async Task Invoke(HttpContext context, ILoggerFactory loggerFactory)
             {
                 await _next(context);
             }
+            catch (BadHttpRequestException ex)
+            {
+                LogException(loggerFactory, ex);
```
tomkerkhove: Let's add a comment here why we are catching these

stijnmoreels: Ok

docs/features/logging.md
```diff
@@ -8,6 +8,9 @@ layout: default
 The `ExceptionHandlingMiddleware` class can be added to the <span>ASP.NET</span> Core pipeline to log unhandled exceptions that are thrown during request processing.
 The unhandled exceptions are caught by this middleware component and are logged through the `ILogger` implementations that are configured inside the project.
 
+The HTTP status code `500` is used as response code when an unhandled exception is caught.
```
tomkerkhove: What about this? It's less technical and tells just what it needs to, no?
```suggestion
The HTTP status code `500` is used as response code when an unhandled exception is caught. However, when the runtime throws a `BadHttpRequestException` we will reflect this by returning the corresponding status code determined by the runtime.
```

docs/features/logging.md
```diff
@@ -8,6 +8,9 @@ layout: default
...
+When an `BadHttpRequestException` is thrown (typically from a built-in <span>ASP.NET</span> component), the `.StatusCode` is used as response code.
```
tomkerkhove: If the above is ok, I'd remove this.

stijnmoreels: Yep, done.

Issue #120, opened 2020-02-27
Support for JWT authorization validation


gverstraete:
**Describe the solution you'd like**
At the moment every project is trying to embrace Managed Identity as much as possible, so as well between Azure API Management & App Service API's.

It would be great to have an AuthorizationFilter which validates on tokens based on a specified property.
We did a similar thing at a project where we are working at the moment. But I hear from others this is also for them interesting.

tomkerkhove:
Certainly interesting - Are you willing to contribute this?

gverstraete:
Yes! But due to time constraints this will be very difficult the upcoming months.

tomkerkhove:
No worries, I'll get this going and you can then review maybe?

tomkerkhove:
Started working on this, here's my checklist:
- [ ] Implement first rough version
- [ ] Provide configuration options
- [ ] Provide documentation
- [ ] Provide more user-friendly way of wiring things up, if possible
- [ ] Provide tests

tomkerkhove:
@fgheysels @gverstraete @stijnmoreels Have you generated JWT tokens on the fly for your unit tests before?

gverstraete:
yes :)

gverstraete:
Here you have a sample which will help you 

public JwtSecurityToken CreateToken(ClaimsPrincipal principal, AuthorizationGrantCodeType type)
{
            var creds = new SigningCredentials(new SymmetricSecurityKey(Encoding.UTF8.GetBytes(_configuration.JwtSigninKey)), SecurityAlgorithms.HmacSha256);
            DateTime expirationTime = default;

            switch (type)
            {
                case AuthorizationGrantCodeType.Bearer:
                    expirationTime = DateTime.Now.AddMinutes(_configuration.BearerExpirationTime);
                    break;
                case AuthorizationGrantCodeType.SAD:
                    expirationTime = DateTime.Now.AddMinutes(_configuration.SadExpirationTime);
                    break;
            }

            return new JwtSecurityToken(
                issuer: _configuration.JwtIssuer,
                audience: _configuration.JwtAudience,
                claims: principal.Claims,
                expires: expirationTime,
                signingCredentials: creds
            );
}

Sorry for the crappy UI; github doesn't seem to like my snippet.

System.IdentityModel.Tokens.Jwt is the namespace

tomkerkhove:
@stijnmoreels will pick up the testing part with real JWT tokens, I will pick up the documentation side of things.

[...]

PR #121, opened 2020-02-27
Provide Azure Managed Identity authorization foundation

tomkerkhove:
Provide Azure Managed Identity authorization which can be used with services such as Azure API Management.

Relates to https://github.com/arcus-azure/arcus.webapi/issues/120

### DIFF
--- a/src/Arcus.WebApi.Security.Authorization/Arcus.WebApi.Security.Authorization.csproj
+++ b/src/Arcus.WebApi.Security.Authorization/Arcus.WebApi.Security.Authorization.csproj
@@ -6,7 +6,7 @@
       <Company>Arcus</Company>
       <RepositoryType>Git</RepositoryType>
       <PackageTags>Azure;WebAPI;App Services;Web App;Web;API</PackageTags>
-      <Description>Provides capabilities to easily build secure Web APIs running in Azure.</Description>
+      <Description>Provides capabilities to easily authorization Web APIs running in Azure.</Description>
       <Copyright>Copyright (c) Arcus</Copyright>
       <PackageLicenseUrl>https://github.com/arcus-azure/arcus.webapi/blob/master/LICENSE</PackageLicenseUrl>
       <PackageProjectUrl>https://github.com/arcus-azure/arcus.webapi</PackageProjectUrl>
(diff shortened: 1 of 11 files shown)


src/Arcus.WebApi.Security.Authorization/Arcus.WebApi.Security.Authorization.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+      <TargetFramework>netstandard2.0</TargetFramework>
+      <Authors>Arcus</Authors>
+      <Company>Arcus</Company>
+      <RepositoryType>Git</RepositoryType>
+      <PackageTags>Azure;WebAPI;App Services;Web App;Web;API</PackageTags>
+      <Description>Provides capabilities to easily build secure Web APIs running in Azure.</Description>
```
tomkerkhove: I'll update this to reflect what the package does

tomkerkhove:
I was going to do it in one PR but agree that it's best to merge this.
Ok for you as well @stijnmoreels?

I can create a preview release and run it at the customer. MyGet package is not an option unfortunately since we're using containers and other NuGet feeds are a pain.

stijnmoreels:
> I was going to do it in one PR but agree that it's best to merge this.
> Ok for you as well @stijnmoreels?
> 
> I can create a preview release and run it at the customer. MyGet package is not an option unfortunately since we're using containers and other NuGet feeds are a pain.

Hmm, I've it's not too much, I would like to keep func/tests together and do it in this PR, no? Not sure what the benifit would be of merging it already?

tomkerkhove:
To start using it early and see if it's decent from a user perspective.

For testing we'll need to wire up APIM or so most probably

stijnmoreels:
> To start using it early and see if it's decent from a user perspective.
> 
> For testing we'll need to wire up APIM or so most probably

Seems a bit quick and dirty to do that on master but for now it's OK I guess.

tomkerkhove:
I've added basic tests for now @stijnmoreels, don't worry - More will follow.

PR #126, opened 2020-03-01
Docs - update feature docs with telemetry serilog correlation enricher

### LINKED ISSUE #68: Provide a Serilog enricher for out-of-the-box correlation information

stijnmoreels:
Closes #68

### DIFF
--- a/docs/features/telemetry.md
+++ b/docs/features/telemetry.md
@@ -1,0 +1,35 @@
+---
+title: "Telemetry Enrichment"
+layout: default
+---
+
+# Telemetry Enrichment
+
+## Serilog Correlation Enrichment
+
+The `Arcus.WebApi.Telemetry.Serilog` library provides a [Serilog enricher](https://github.com/serilog/serilog/wiki/Enrichment) 
+that adds the correlation information of the current request to the log event as a log property called `TransactionId` and `OperationId`.
+
+**Example**
+
+- `TransactionId`: `A5E90591-ADB0-4A56-818A-AC5C02FBFF5F`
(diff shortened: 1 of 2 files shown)


docs/features/telemetry.md
```diff
@@ -0,0 +1,35 @@
...
+
+The enricher requires access to the application services so it can get the correlation information.
+Following example shows how the Serilog logger is configured in the `Startup.cs` file.
+
+```csharp
+public class Startup
+{
+    public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+    {
+        Log.Logger = new LoggerConfiguration()
```
tomkerkhove: We tend to also configure them on startup, see template.

Either we have to align this approach in template, which is ok, or also doc how to do it on startup which is harder I think.

What I've seen is that people tend to do 2 things:
1. Create small logger for startup
2. Create more detailed logger during configuration with all the bells and whistles

If that's ok for you, then we should align in our template and create an issue there.

stijnmoreels: Yes, I agree to change it in the template.

stijnmoreels: Done: https://github.com/arcus-azure/arcus.templates/issues/114

PR #129, opened 2020-03-03
Provide configuration options for Managed Identity authorization validation

tomkerkhove:
Provide configuration options for Managed Identity authorization validation.

Next PR will finalize this with docs, registration and JWT tests.

Relates to #120

### DIFF
--- a/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationFilter.cs
+++ b/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationFilter.cs
@@ -30,8 +30,9 @@ namespace Arcus.WebApi.Security.Authorization
         public virtual async Task OnAuthorizationAsync(AuthorizationFilterContext context)
         {
             var jwtString = context.HttpContext.Request.Headers[_headerName].ToString();
+            var isValidToken = await _jwtTokenReader.IsValidTokenAsync(jwtString);
 
-            if (string.IsNullOrWhiteSpace(jwtString) || !await _jwtTokenReader.IsValidTokenAsync(jwtString))
+            if (string.IsNullOrWhiteSpace(jwtString) || isValidToken == false)
             {
                 context.Result = new UnauthorizedObjectResult("Unauthorized because of wrong JWT MSI token.");
             }
--- a/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationFilter.cs
+++ b/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationFilter.cs
@@ -31,7 +31,7 @@ namespace Arcus.WebApi.Security.Authorization
         {
             var jwtString = context.HttpContext.Request.Headers[_headerName].ToString();
 
-            if (string.IsNullOrWhiteSpace(jwtString) || !await _jwtTokenReader.IsValidToken(jwtString))
+            if (string.IsNullOrWhiteSpace(jwtString) || !await _jwtTokenReader.IsValidTokenAsync(jwtString))
             {
                 context.Result = new UnauthorizedObjectResult("Unauthorized because of wrong JWT MSI token.");
             }
--- a/src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
+++ b/src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
@@ -15,7 +15,7 @@ namespace Arcus.WebApi.Security.Authorization.Jwt
 
         private readonly string _applicationId;
         private readonly TokenValidationParameters _tokenValidationParameters;
-        private readonly JwtSecurityTokenHandler _handler;
+        private readonly JwtSecurityTokenHandler _handler = new JwtSecurityTokenHandler();
         private readonly ConfigurationManager<OpenIdConnectConfiguration> _configManager;
 
         /// <summary>
@@ -23,7 +23,7 @@ namespace Arcus.WebApi.Security.Authorization.Jwt
         /// </summary>
         /// <remarks>Uses Microsoft OpenId connect discovery endpoint</remarks>
         /// <param name="applicationId">Azure AD Application used as audience to validate against</param>
-        public JwtTokenReader(string applicationId) : this()
+        public JwtTokenReader(string applicationId)
         {
             Guard.NotNullOrWhitespace(applicationId, nameof(applicationId));
 
@@ -45,19 +45,13 @@ namespace Arcus.WebApi.Security.Authorization.Jwt
         /// </summary>
         /// <param name="openIdConnectDiscoveryUri">Uri of an OpenId connect endpoint for discovering the configuration</param>
         /// <param name="tokenValidationParameters">Collection of parameters to influence how the token validation is done</param>
-        public JwtTokenReader(TokenValidationParameters tokenValidationParameters, string openIdConnectDiscoveryUri) :
-            this()
+        public JwtTokenReader(TokenValidationParameters tokenValidationParameters, string openIdConnectDiscoveryUri)
         {
             Guard.NotNullOrWhitespace(openIdConnectDiscoveryUri, nameof(openIdConnectDiscoveryUri));
             Guard.NotNull(tokenValidationParameters, nameof(tokenValidationParameters));
 
             _tokenValidationParameters = tokenValidationParameters;
             _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(openIdConnectDiscoveryUri, new OpenIdConnectConfigurationRetriever());
-        }
-
-        private JwtTokenReader()
-        {
-            _handler = new JwtSecurityTokenHandler();
         }
 
         /// <summary>
--- a/src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
+++ b/src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
@@ -71,9 +71,10 @@ namespace Arcus.WebApi.Security.Authorization.Jwt
                 ValidateLifetime = true
             };
 
+            SecurityToken jwtToken;
             try
             {
-                _handler.ValidateToken(token, validationParameters, out _);
+                _handler.ValidateToken(token, validationParameters, out jwtToken);
             }
             catch (Exception)
             {
--- a/src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
+++ b/src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
@@ -14,18 +14,50 @@ namespace Arcus.WebApi.Security.Authorization.Jwt
             "https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration";
 
         private readonly string _applicationId;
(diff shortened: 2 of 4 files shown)


tomkerkhove:
@fgheysels @gverstraete @stijnmoreels Have you generated JWT tokens on the fly for your unit tests before?

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -36,7 +68,7 @@ public async Task<bool> IsValidToken(string token)
         {
```
fgheysels: Should async methods be suffixed with Async ?  Or don't we do that within arcus ?

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -14,18 +14,50 @@ public class JwtTokenReader : IJwtTokenReader
...
+        public JwtTokenReader(TokenValidationParameters tokenValidationParameters) : this(MicrosoftDiscoveryEndpoint, tokenValidationParameters)
+        {
+        }
 
+        /// <summary>
+        ///     Constructor
+        /// </summary>
+        /// <param name="openIdConnectDiscoveryUri">Uri of an OpenId connect endpoint for discovering the configuration</param>
+        /// <param name="tokenValidationParameters">Collection of parameters to influence how the token validation is done</param>
+        public JwtTokenReader(string openIdConnectDiscoveryUri, TokenValidationParameters tokenValidationParameters) :
```
fgheysels: Don't know if it makes sense to swap the parameters on this ctor ?  What's your idea on this @stijnmoreels ?

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -14,18 +14,50 @@ public class JwtTokenReader : IJwtTokenReader
             "https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration";
 
         private readonly string _applicationId;
+        private readonly TokenValidationParameters _tokenValidationParameters;
```
fgheysels: If I'm not mistaken, the AudienceId is also present in the `TokenValidationParameters` , so in fact, you have 2 ways here to identify the audience (the property in `TokenValidationParametes` and the `_applicationId`).

Shouldn't the ctor that accepts the `applicationId` not instantiate a `TokenValidationParameters` instance, fill out the `ValidAudience` property and specify some defaults for the other properties ?
Then the `_applicationId`member can be removed, as this only adds confusion.

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
tomkerkhove: Good idea

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
tomkerkhove: Yes, but it's enforced by `IJwtTokenReader`

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
fgheysels: I see that you do the above  in the `ValidateToken` method; I'd do that in the ctor, and then work with the `TokenValidationParameters` in the ValidateToken method.

tomkerkhove: That's the idea but the default uses the signing keys from the remote endpoint which requires async.
That's why I decided to defer it and use this approach: https://github.com/arcus-azure/arcus.webapi/pull/129/files/035318cff93aac825afbd56f6c7eb101455994cd#diff-afcdc5e9bb40ea179ba752371ed1d6c8R71

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
stijnmoreels: Yes, I agree with @fgheysels . It's better to gradually show overloads with the same parameters in the beginning. Better discovery.
But yeah, already handled and fixed, so what am I doing here 🤡 .

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -14,18 +14,50 @@ public class JwtTokenReader : IJwtTokenReader
...
+            this()
+        {
+            Guard.NotNullOrWhitespace(openIdConnectDiscoveryUri, nameof(openIdConnectDiscoveryUri));
+            Guard.NotNull(tokenValidationParameters, nameof(tokenValidationParameters));
+
+            _tokenValidationParameters = tokenValidationParameters;
+            _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(openIdConnectDiscoveryUri, new OpenIdConnectConfigurationRetriever());
+        }
+
+        private JwtTokenReader()
```
stijnmoreels: I find it a bit tricky to guide every ctor overload to this overload, just to set the field. It's easily forgotton I guess.
If it's needed anyway, why not make it static?

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
tomkerkhove: Already done ;)

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
tomkerkhove: Good idea

fgheysels: I'm in favor of initializing it outside of the ctor.  However, don't know if it can be static.   It depends a bit on what this JwtSecurityHandler instance does.   Does it keep state ?  In that case, it is tricky to declare it as static imo.

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
fgheysels: True, but the `IJwtTokenReader` is owned by this library as well ?  (It is defined here).

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
fgheysels: Ok , I see.

But then, why not use a Factory method instead of this constructor:

```csharp
public static async Task<JwtTokenReader> CreateAsync( string applicationId )
{
     OpenIdConnectConfiguration config = await _configManager.GetConfigurationAsync();

     var validationParameters = new TokenValidationParameters	 
     {	            
         ValidateAudience = true,
         ValidAudience = applicationId,	 
         ValidateIssuer = false,	                
         IssuerSigningKeys = config.SigningKeys,
         ValidateLifetime = true	                
     };

     return new JwtTokenReader(validationParameters);
}
```
something like this.

Just my 2 cents.  Then offcourse, this is not very much inline with the other options to instantiate a JwtTokenReader, since there, constructors are used.

Anyway, since `_applicationId` is a private member, and this is just internal workings, I can agree with how it is implemented.

/writing-while-thinking

tomkerkhove: That's certainly another approach but problem is that startup is not async as well so you'll have the same issue unless there is a way which I don't know of course!

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
tomkerkhove: I'll initialize it on the field without static!

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
tomkerkhove: Ahaha I'm a moron 😂 Yes, I'll change it

src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationFilter.cs
```diff
@@ -31,7 +31,7 @@ public virtual async Task OnAuthorizationAsync(AuthorizationFilterContext contex
         {
             var jwtString = context.HttpContext.Request.Headers[_headerName].ToString();
 
-            if (string.IsNullOrWhiteSpace(jwtString) || !await _jwtTokenReader.IsValidToken(jwtString))
+            if (string.IsNullOrWhiteSpace(jwtString) || !await _jwtTokenReader.IsValidTokenAsync(jwtString))
```
stijnmoreels: I'm fine with inline `await` in `if`, and inverted condition. But maybe too much for Arcus starter devs/users?

src/Arcus.WebApi.Unit/Security/Authorization/JwtTokenReaderTests.cs
```diff
@@ -28,5 +29,53 @@ public void Constructor_WithApplicationId_Succeeds()
             // Assert
             Assert.NotNull(jwtTokenReader);
         }
+
```
stijnmoreels: Is there a way to test the validation token more thoroughly?

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -45,10 +71,9 @@ public async Task<bool> IsValidToken(string token)
                 ValidateLifetime = true
             };
 
-            SecurityToken jwtToken;
             try
             {
-                _handler.ValidateToken(token, validationParameters, out jwtToken);
+                _handler.ValidateToken(token, validationParameters, out _);
```
stijnmoreels: I guess for Arcus code rules we should provide a value 😄 .

src/Arcus.WebApi.Unit/Security/Authorization/JwtTokenReaderTests.cs
stijnmoreels: Like, maybe generate a valid token (--> pass), and generate a token and alter some values (--> fail).

tomkerkhove: That's mentioned on issue and why it's not closed yet, more to come

src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationFilter.cs
tomkerkhove: It's not related to starter devs/users, it's just that this is less readible so I've updated this indeed.

Forgot to do that, sorry!

src/Arcus.WebApi.Security.Authorization/Jwt/JwtTokenReader.cs
tomkerkhove: Discussed this with Frederik and he's ok with the current approach although it's not ideal, but nothing we can do about it

PR #133, opened 2020-03-30
Use Arcus Security 1.1

### LINKED ISSUE #90: Use Arcus Security v1.x

tomkerkhove:
Closes #90


src/Arcus.WebApi.Tests.Unit/Security/InMemorySecretProvider.cs
```diff
@@ -3,6 +3,7 @@
 using System.Linq;
```
stijnmoreels: I guess there are some methods that can be removed?

tomkerkhove: Done

Issue #115, opened 2020-02-06
Introduce Request logging middleware


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
In certain scenarios you'd like to log every inbound request. It would be ideal if we provide middleware that does this out-of-the-box.

**Describe the solution you'd like**
We can introduce a `RequestTrackingMiddleware` that uses https://github.com/arcus-azure/arcus.observability/issues/14.

Things to keep in mind:
- Configuration should be provided in what should be logged:
    - Headers only, body only, headers and body, none of both, etc.
- We should provide sanitization out-of-the-box, with possibility to provide custom sanitization (based on inheritance)

**Describe alternatives you've considered**
I remember looking at [Serilog.AspNetCore's Request logging](https://github.com/serilog/serilog-aspnetcore#request-logging) which didn't feel good enough for me but can't remember why.

I think it was because it didn't work nicely with Azure Application Insights but I'm not sure.

stijnmoreels:
> Describe alternatives you've considered
>I remember looking at Serilog.AspNetCore's Request logging which didn't feel good enough for me but can't remember why
>
> I think it was because it didn't work nicely with Azure Application Insights but I'm not sure.
Maybe because the request logging only logs the after the response is created. And I think it's rather minimal.

tomkerkhove:
That's ok, but it's more because it logs it as a trace and not AI request which is not ideal if you ask me

tomkerkhove:
https://github.com/arcus-azure/arcus.observability/pull/35 is now merged and work can be started.

stijnmoreels:
Maybe besides configuring "headers" or "no headers", also which headers. Bc, security headers with secret values is maybe not a good idea to log in plain text.

tomkerkhove:
We should do this indeed but via this approach:
> We should provide sanitization out-of-the-box, with possibility to provide custom sanitization (based on inheritance)

So it's either headers or no headers, if headers are added then our sanitization will remove documented headers out-of-the-box; but I would not make it configurable as they can extend/override our sanitization

PR #140, opened 2020-04-13
Feature - add request tracking middleware

stijnmoreels:
Relates to #115


docs/features/logging.md
```diff
@@ -1,29 +1,39 @@
 ---
-title: "Logging unhandled exceptions via ASP.NET Core middleware"
+title: "Logging via ASP.NET Core middleware"
```
tomkerkhove, line 2: Maybe just call it logging?
```suggestion
title: "Logging"
```

docs/features/logging.md
```diff
@@ -45,7 +109,7 @@ WebHost.CreateDefaultBuilder(args)
 
 To be able to use the `AddApplicationInsights` extension method, the Microsoft.Extensions.Logging.ApplicationInsights package must be installed.
 
-
 If the parameter-less `AddApplicationInsights` method is used, the configurationsetting `ApplicationInsights:InstrumentationKey` must be specified and the value of the instrumentation-key of the Application Insights resource must be set.
 
+
```
tomkerkhove, line 114:
```suggestion
```

docs/features/logging.md
```diff
@@ -32,9 +42,63 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
+        return new Dictionary<string, object>();
+    }
+}
+```
+
+So the resulting log message becomes:
+
+`HTTP Request POST http://localhost:5000/weatherforecast completed with 200 in 00:00:00.0191554 at 03/23/2020 10:12:55 +00:00 - (Context: [Content-Type, application/json])`
+
+## Application Insights
```
tomkerkhove, line 99: Frankly, I think we should recommend using Serilog as well since the requests will be tracked as "traces" when using the current suggestion here.

Would you mind adding a second option?

docs/index.md
```diff
@@ -31,7 +31,7 @@ For more granular packages we recommend reading the documentation.
 - **Telemetry**
     - [Enrich with telemetry information](features/telemetry)
 - **Logging**
-    - [Logging unhandled exceptions](features/logging)
+    - [Logging](features/logging)
```
tomkerkhove, line 34: This is duplicate from the header, let's add an entry for both features to improve discoverability

src/Arcus.WebApi.Correlation/HttpCorrelationInfoAccessor.cs
```diff
@@ -27,7 +27,7 @@ public HttpCorrelationInfoAccessor(IHttpContextAccessor contextAccessor)
         /// </summary>
         public CorrelationInfo GetCorrelationInfo()
         {
-            var correlationInfo = _httpContextAccessor.HttpContext.Features.Get<CorrelationInfo>();
+            var correlationInfo = _httpContextAccessor.HttpContext?.Features.Get<CorrelationInfo>();
```
tomkerkhove, line 30: Aren't we adding a potential null pointer here? Shouldn't we check if features is null as well?

src/Arcus.WebApi.Tests.Unit/Hosting/TestApiServer.cs
```diff
@@ -112,18 +120,48 @@ protected override void ConfigureWebHost(IWebHostBuilder builder)
...
-            builder.UseStartup<TestStartup>();
+            builder.Configure(app =>
+            {
+                foreach (Action<IApplicationBuilder> configure in _configures)
+                {
+                    configure(app);
+                }
+
+                app.UseMiddleware<ExceptionHandlingMiddleware>();
+                app.UseMiddleware<TraceIdentifierMiddleware>();
```
tomkerkhove, line 131: What is this? 🤔

src/Arcus.WebApi.Tests.Unit/Hosting/TestApiServer.cs
```diff
@@ -112,18 +120,48 @@ protected override void ConfigureWebHost(IWebHostBuilder builder)
...
+
+                app.UseHttpCorrelation();
+                app.UseSerilogRequestLogging();
```
tomkerkhove, line 134: We don't need that as that's what we are adding

src/Arcus.WebApi.Tests.Unit/Logging/RequestTrackingMiddlewareTests.cs
```diff
@@ -0,0 +1,130 @@
...
+                }
+            }
+        }
+
+        [Fact]
+        public async Task GetRequestWithoutHeaders_TracksRequest_ReturnsSuccess()
+        {
+            // Arrange
+            const string headerName = "x-custom-header", headerValue = "custom header value", body = "echo me";
+            _testServer.AddConfigure(app => app.UseMiddleware<NoHeadersRequestTrackingMiddleware>());
```
tomkerkhove, line 56: I'm not sure what we are trying to test here? Can you elaborate on the scenario please?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,118 @@
...
+                await TrackRequestAsync(httpContext, stopwatch.Elapsed);
+            }
+        }
+
+        private async Task TrackRequestAsync(HttpContext httpContext, TimeSpan duration)
+        {
+            try
+            {
+                var headers = ExtractRequestHeaders(httpContext.Request.Headers) ?? new Dictionary<string, object>();
+                var body = await ExtractRequestBodyAsync(httpContext.Request.Body) ?? new Dictionary<string, object>();
```
tomkerkhove, line 61: I'm thinking this would be better to move to `LogRequest` as an optional parameter which is off by default, what do you think? For example, `includeRequestBody` or so?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+
+namespace Arcus.WebApi.Logging
+{
+    /// <summary>
+    /// Request tracing middleware component to log every incoming HTTP request.
+    /// </summary>
+    public class RequestTrackingMiddleware
```
tomkerkhove, line 15: I'm missing the configuration aspect a bit from the issue:

> Configuration should be provided in what should be logged:
> - Headers only, body only, headers and body, etc.

Can we provide an options/configuration approach similar to other flow?

docs/features/logging.md
```diff
@@ -32,9 +42,63 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
-If you have multiple middleware components configured, it is advised to put the `ExceptionHandlingMiddleware` as the first one.  By doing so, unhandled exceptions that might occur in other middleware components will also be logged by the `ExceptionHandlingMiddleware` component.
+If you have multiple middleware components configured, it is advised to put the `ExceptionHandlingMiddleware` as the first one.
+By doing so, unhandled exceptions that might occur in other middleware components will also be logged by the `ExceptionHandlingMiddleware` component.
+
+## Logging incoming requests
+
+The `RequestTrackingMiddleware` class can be added to the <span>ASP.NET</span> Core pipeline to log all received HTTP requests.
+The requests by this middleware component are logged through the `ILogger` implementations that are configured inside the project.
+
+The HTTP request headers and request body are by default logged.
```
tomkerkhove, line 53: We shouldn't log these by default from a security perspective imo or should add a variety of default headers to omit

docs/features/logging.md
```diff
@@ -32,9 +42,63 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
+
+    ...
+    app.UseMvc();
+}
+```
+
+**Configuration**
 
-To get the exceptions logged in Azure Application Insights, configure logging like this when building the `IWebHost` in `Program.cs`:
+Optionally, one can inherit from this middleware component and override the default request header and body extraction to discared one or both.
```
tomkerkhove, line 75: This should only be done for sanitization imo or if you prefer an action for sanitization that's also an option

src/Arcus.WebApi.Correlation/HttpCorrelationInfoAccessor.cs
stijnmoreels: That's a list and is in practice never `null`, but we could, yeah.

src/Arcus.WebApi.Tests.Unit/Hosting/TestApiServer.cs
stijnmoreels: Oh, that's a nice one, that's test middleware to enable/disable the `Request.TraceIdentifier` from the HTTP request so we can see that the correlation middleware correctly generates one for us if it's not present.

src/Arcus.WebApi.Tests.Unit/Hosting/TestApiServer.cs
stijnmoreels: Oh, yes.

src/Arcus.WebApi.Tests.Unit/Logging/RequestTrackingMiddlewareTests.cs
stijnmoreels: Yes, the added `.AddConfigure()` method is a way to provide additional middleware components from the test. In this case, we want to verify if we can override the existing middleware and provide one of our own (in this case `NoHeadersRequestTrackingMiddleware`) and that indeed the "request headers are not" in the request tracking.

stijnmoreels: The three tests are the three useful combinations of the available overridables. We could add also where both the headers and body is not present though.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: No those are present, you can override the middleware like the issue described was the way to go.

docs/features/logging.md
stijnmoreels: Was indeed also my concern in the issue. Was not sure how it was eventually resolved.

docs/features/logging.md
stijnmoreels: You mean that the user should get the opportiunity to sanitize the body? That can be done with overriden method.

docs/index.md
stijnmoreels: Yes.

stijnmoreels:
But given the questions and comments on this PR I figure that I've been misguided by the issue and this is not the way you wanted things to be implemented (it's actually not entirely like I wanted too).
So, after you looked at the `RequestTrackingMiddleware` you can give me some directions in what ways I do have to implement this tracking middleware.

docs/features/logging.md
```diff
@@ -1,29 +1,39 @@
...
 
-The `ExceptionHandlingMiddleware` class can be added to the <span>ASP.NET</span> Core pipeline to log unhandled exceptions that are thrown during request processing.
-The unhandled exceptions are caught by this middleware component and are logged through the `ILogger` implementations that are configured inside the project.
+The `Arcus.WebApi.Logging` package provides a way to log several kinds of information during the receival and handling of HTTP requests.
+
+- [Logging unhandled exceptions](#logging-unhandled-exceptions)
+- [Logging incoming requests](#logging-incoming-requests)
 
-The HTTP status code `500` is used as response code when an unhandled exception is caught. However, when the runtime throws a `BadHttpRequestException` we will reflect this by returning the corresponding status code determined by the runtime.
+To send the logging information to Application Insights, see [this explenation](#application-insights).
```
tomkerkhove, line 13:
```suggestion
To send the logging information to Application Insights, see [this explanation](#application-insights).
```

docs/features/logging.md
tomkerkhove: Let's not do this then, what do you think about my suggestion?

docs/features/logging.md
tomkerkhove: Sure, but what do you mean with this?
> override the default request header

src/Arcus.WebApi.Tests.Unit/Hosting/TestApiServer.cs
tomkerkhove: So it's our own internal stuff or not?

stijnmoreels: Yes, test middleware to manipulate the `Request.TraceIdentifier`.

docs/features/logging.md
stijnmoreels: I think moving it to the log request is an option for the headers, but maybe have an default list of headers to look for (like `Accept`, `Content-Type`, ...), but the body is a different issue bc it can be serialized in diff ways. 
Either way, just adding two booleans is not really something I would vote for.

stijnmoreels: Since we're in Arcus and passing functions is not an option, we could create/use an serialization interface for this.

docs/features/logging.md
stijnmoreels: That should be `override the default request headers`.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: Inheritance should only be required if you want to sanitize, not for those things as per the issue:

> Things to keep in mind:
> 
> - Configuration should be provided in what should be logged:
>    - Headers only, body only, headers and body, etc.
> - We should provide sanitization out-of-the-box, with possibility to provide custom sanitization *(based on inheritance)*

docs/features/logging.md
tomkerkhove: I'm still not sure what we mean with this? 🤔

fgheysels: I think that the meaning is that you can inherit from this middleware, add your own functionality which allows you to remove one or more headers that are in the incoming request (or overwrite them).  (But why would you do that ? )

docs/features/logging.md
```diff
@@ -1,29 +1,39 @@
...
+- [Logging unhandled exceptions](#logging-unhandled-exceptions)
+- [Logging incoming requests](#logging-incoming-requests)
 
-The HTTP status code `500` is used as response code when an unhandled exception is caught. However, when the runtime throws a `BadHttpRequestException` we will reflect this by returning the corresponding status code determined by the runtime.
+To send the logging information to Application Insights, see [this explanation](#application-insights).
 
 ## Installation
 
-This feature requires to install our NuGet package
+These features requires to install our NuGet package
```
fgheysels, line 17: require instead of requires

docs/features/logging.md
tomkerkhove: For security reasons you'd want to have the option to truncate headers for sure.

Oh now I get it

docs/features/logging.md
```diff
@@ -32,9 +42,63 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
 
-If you have multiple middleware components configured, it is advised to put the `ExceptionHandlingMiddleware` as the first one.  By doing so, unhandled exceptions that might occur in other middleware components will also be logged by the `ExceptionHandlingMiddleware` component.
+If you have multiple middleware components configured, it is advised to put the `ExceptionHandlingMiddleware` as the first one.
+By doing so, unhandled exceptions that might occur in other middleware components will also be logged by the `ExceptionHandlingMiddleware` component.
+
+## Logging incoming requests
 
-To get the exceptions logged in Azure Application Insights, configure logging like this when building the `IWebHost` in `Program.cs`:
+The `RequestTrackingMiddleware` class can be added to the <span>ASP.NET</span> Core pipeline to log all received HTTP requests.
+The requests by this middleware component are logged through the `ILogger` implementations that are configured inside the project.
```
fgheysels, line 51: I'd suggest to rephrase this to
> The incoming requests are logged by this middleware component by the `ILogger` implementations that are configured in the project

docs/features/logging.md
fgheysels: I think it'd be good if there's a list of predefined header names to look for which to omit from the logs, or -maybe a better approach- sanitze the values of those 'sensitive' headers.
(I'm thinking on headers like `Authentication`  , apikeys, etc...

src/Arcus.WebApi.Tests.Unit/Logging/NoHeadersRequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,34 @@
...
+            ILogger<RequestTrackingMiddleware> logger) 
+            : base(next, logger)
+        {
+        }
+
+        /// <summary>
+        /// Extracts information from the given HTTP <paramref name="requestHeaders"/> to include in the request tracking context.
+        /// </summary>
+        /// <param name="requestHeaders">The headers of the current HTTP request.</param>
+        protected override IDictionary<string, object> ExtractRequestHeaders(IHeaderDictionary requestHeaders)
```
tomkerkhove, line 29: Let's keep this, but maybe call it this:
```suggestion
        protected override IDictionary<string, object> SanitizeRequestHeaders(Dictionary requestHeaders)
```

Next to that, we should support providing a list of headers to remove in the configuration to avoid having to do this.

src/Arcus.WebApi.Tests.Unit/Logging/NoBodyRequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,34 @@
...
+        public NoBodyRequestTrackingMiddleware(RequestDelegate next, ILogger<RequestTrackingMiddleware> logger)
+            : base(next, logger)
+        {
+        }
+
+        /// <summary>
+        /// Extracts information from the given HTTP request body <paramref name="requestStream"/> to include in the request tracking context.
+        /// </summary>
+        /// <param name="requestStream">The body of the current HTTP request.</param>
+        protected override Task<IDictionary<string, object>> ExtractRequestBodyAsync(Stream requestStream)
```
tomkerkhove, line 29: As discussed, we will not do this for now

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: As discussed, people will be able to:
- Sanitize headers by providing a list of sensitive headers to remove; and/or by using inheritance
- What needs to be logged will be made configurable (headers only, body only, none which means just URL)

docs/features/logging.md
tomkerkhove: I like the idea @fgheysels, we'll add a default list; what would be your suggestions?

- `Authentication`
- `X-API-Key`

Anything else? We can still add more in the future ofcourse.

src/Arcus.WebApi.Tests.Unit/Logging/NoHeadersRequestTrackingMiddleware.cs
stijnmoreels: That's done.

docs/features/logging.md
stijnmoreels: Done that.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: Done that.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: We could provide the same options as in the `RequestTrackingOptions` with a series of by default omitted headers and two booleans for headers/body? 🤔

docs/features/logging.md
fgheysels: Every header with `key` in the name ?

stijnmoreels: I also added the request header for the client certificate in app services.

src/Arcus.WebApi.Logging/RequestTrackingOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// Gets or sets the value to indicate whether or not the HTTP request body should be tracked.
+        /// </summary>
+        public bool IncludeRequestBody { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the HTTP request headers names that will be omitted during request tracking.
+        /// </summary>
+        public ICollection<string> OmittedHeaderNames { get; set; } = new Collection<string> { "Authentication", "X-Api-Key", "X-ARR-ClientCert" };
```
stijnmoreels, line 24: It's not really recommended to expose the collection directly in classes, but in this case with an options type, it's better I think. No?🤔

docs/features/logging.md
stijnmoreels: I changed it to a Serlog example.

src/Arcus.WebApi.Logging/RequestTrackingOptions.cs
stijnmoreels: Made it to a collection so you could easily add one.

docs/features/logging.md
```diff
@@ -32,20 +42,123 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
+If you have multiple middleware components configured, it is advised to put the `ExceptionHandlingMiddleware` as the first one.
+By doing so, unhandled exceptions that might occur in other middleware components will also be logged by the `ExceptionHandlingMiddleware` component.
+
+## Logging incoming requests
+
+The `RequestTrackingMiddleware` class can be added to the <span>ASP.NET</span> Core pipeline to log all received HTTP requests.
+The incoming requests are logged by this middleware component using the `ILogger` implementations that are configured in the project.
+
+The HTTP request headers are logged by default, except certain security headers are by default omitted: `Authentication`, `X-Api-Key` and `X-ARR-ClientCert`.
+The HTTP request body is not logged by default.
```
tomkerkhove, line 54: I'd move this on line 53 as well and add a new line here for formatting reasons

docs/features/logging.md
```diff
@@ -32,20 +42,123 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
+        // (default: `false`)
+        options.IncludeRequestBody = true;
+
+        // Whether or not the configured HTTP request headers should be included in the request tracking logging emits.
+        // (default: `true`)
+        options.IncludeRequestHeaders = true;
+
+        // All omitted HTTP request header names that should always be excluded from the request tracking logging emits.
+        // (default: `[ "Authentication", "X-Api-Key", "X-ARR-ClientCert" ]`)
+        options.OmittedRequestHeaderNames.Add("X-My-Secret-Header");
```
tomkerkhove, line 94: Question - Do we have our defaults in this list? If not, it would be ideal so that folks can remove them if they want them to be included (on their responsibility of course)

docs/features/logging.md
```diff
@@ -32,20 +42,123 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
+    public EmptyButNotOmitRequestTrackingMiddleware(
+        RequestDelegate next,
+        ILogger<RequestTrackingMiddleware> logger) 
+        : base(next, logger)
+        {
+        }
+
+    protected override IDictionary<string, StringValues> SanitizeRequestHeaders(IDictionary<string, StringValues> requestHeaders)
+    {
+        requestHeaders["X-Api-Key"] = String.Empty;
```
tomkerkhove, line 118: Like this or just call remove on it? Anyhow, it's just a sample

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            try
+            {
+                IDictionary<string, StringValues> headers = 
+                    _options.IncludeRequestHeaders
+                        ? SanitizeRequestHeaders(httpContext.Request.Headers) ?? new Dictionary<string, StringValues>()
+                        : new Dictionary<string, StringValues>();
+                
+                
+                IDictionary<string, StringValues> body =
```
tomkerkhove, line 77: This feels more readible to me personally:
```csharp
IDictionary<string, StringValues> body = new Dictionary<string, StringValues>();
if(_options.IncludeRequestBody)
{
  body = await SanitizeRequestBodyAsync(httpContext.Request.Body);
}
```

tomkerkhove: Same for headers

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Extracts information from the given HTTP <paramref name="requestHeaders"/> to include in the request tracking context.
+        /// </summary>
+        /// <param name="requestHeaders">The headers of the current HTTP request.</param>
+        protected virtual IDictionary<string, StringValues> SanitizeRequestHeaders(IDictionary<string, StringValues> requestHeaders)
+        {
+            return requestHeaders.Where(header => !_options.OmittedHeaderNames.Contains(header.Key));
```
tomkerkhove, line 97: But this is just personal preference, you can ignore if you want
```suggestion
            return requestHeaders.Where(header => _options.OmittedHeaderNames.Contains(header.Key) == false);
```

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            if (requestStream.CanSeek)
+            {
+                if (requestStream.Position > 0)
+                {
+                    requestStream.Position = 0;
+                }
+            }
+
+            using (var reader = new StreamReader(requestStream))
```
tomkerkhove, line 115: Can we dispose this? What happens if controller reads it again? Is that what the buffering option covers?

src/Arcus.WebApi.Logging/RequestTrackingOptions.cs
tomkerkhove: I think this is ok!

docs/features/logging.md
stijnmoreels: Yes, ok.

docs/features/logging.md
stijnmoreels: Yes, the defaults are already in the `.OmittedRequestHeaderNames`. Add/Remove is now made easy, but also removing them altogheter by overriding with a new collection.

docs/features/logging.md
stijnmoreels: Well, the example here is to show that you make some custom logic during sanitization of the headers. If we would call remove, we could have done it in the options by calling `.OmittedRequestHeaderNames.Remove("X-Api-Key")`, but this example tries to show that you can run something else: not removing, but clearing the values.

But indeed, just for demonstration purposes. I thought that clearing values would also be a good thing to have bc you can see in the logs if a header was added without exposing secret information.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: Yes, I thought it would 😄 . For me it's the other way arround, I guess that's just functional programming genes. I'll change it for the Arcus consumers.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: Hmm, 🤔 , flipping booleans is less work than a new condition. But I see that people would like this more. I'll change it for you. But again, something I would normally not do.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: Yes, the reading inside the controller comes first, actually, (during the `await _next(httpContext)` call), but indeed after that we have to read it again here. That's indeed where the buffering comes into play.

stijnmoreels: Except if there's another middleware component after this one that also needs to read the request body, I don't see a problem 🤔 . It may get disposed anyway after this request middleware, but you can argue if it's the middleware's job to do so.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: Oh you didn't have to but ok; thanks I guess!

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: `!` just sucks from a readabilty perspective - You need to clearly look for it to notice it in my opinion; but maybe I'm reviewing too much lately 😂

docs/features/logging.md
tomkerkhove: Oh I get it now, thought it was to remove it!

Maybe this proves more what we want to show?

```suggestion
        requestHeaders["X-Api-Key"] = "<redacted>";
```

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: Perfect, we're good to go then!

stijnmoreels:
> LGTM!

I'll merge after build's green?

stijnmoreels:
Some final remarks, @fgheysels ?

docs/features/logging.md
stijnmoreels: Monkey, keystroke, donkey, keyboard, whiskey, jockey, smokey, ?😂

fgheysels: That would indeed be to restricting.   I think we can start with the 2 headers that Tom mentions and see if we need to extend it later

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,134 @@
...
+                return new Dictionary<string, StringValues>();
+            }
+
+            if (requestStream.CanSeek)
+            {
+                if (requestStream.Position > 0)
+                {
+                    requestStream.Position = 0;
+                }
+            }
```
fgheysels, line 122: I would say that we should not log the body if the stream doesn't support seeking ?
In that case (non seekable stream), you won't be able to reset it to it's initial position.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            using (var reader = new StreamReader(requestStream))
+            {
+                string contents = await reader.ReadToEndAsync();
```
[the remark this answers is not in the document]
stijnmoreels: Not by default, no, that's why a we use `.EnableBuffering()` when the body sanitization is enabled.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: That could be an extra fail-safe, yes, bus as I told here above, using `.EnabledBuffering()` will hold the request body in memory and will write it on disk when it exceeds a given threshold.

stijnmoreels: On the other hand, the body can not be seekable, but could be set on position `0`.

stijnmoreels: Voila, [changed it](https://github.com/arcus-azure/arcus.webapi/pull/140/files#diff-dd8dfb1d0fb4a65ce66a770f805d082bR124), OK for you? @fgheysels @tomkerkhove ?

tomkerkhove: > On the other hand, the body can not be seekable, but could be set on position `0`.

AFAIK that won't work as you are "seeking" it to 0

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,148 @@
...
+                catch
+                {
+                    return new Dictionary<string, StringValues>();
+                }
+            }
+
+            using (var reader = new StreamReader(requestStream))
+            {
+                string contents = await reader.ReadToEndAsync();
+                return new Dictionary<string, StringValues>
```
tomkerkhove, line 141: This still feels a bit odd to me, think this one should just return the raw string and you can then move it to a dictionary if you want; but from a consumer perspective if you are overriding this you shouldn't have to care about this (that's our problem)

stijnmoreels: Ah, OK, I thought that it was also a consumer plus to have the opportiunity to add additional entries, or even split your body into separate entries. But I'll update this then to just returning the raw body.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
fgheysels: The position is only set to zero if the stream is seekable, so that's ok @tomkerkhove ?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -0,0 +1,144 @@
...
+                try
+                {
+                    if (requestStream.Position != 0)
+                    {
+                        return null;
+                    }
+                }
+                catch
+                {
+                    return null;
```
fgheysels, line 133: Which exceptions do we expect here ?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: I was merely explaining why but didn't check the code

PR #147, opened 2020-04-15
🚧 Design Review - Providing easy way to wire-up the Azure Managed Identity authorization

tomkerkhove:
Design Review for providing easy way to wire-up the Azure Managed Identity authorization.

What do you think of this approach?

Normally we use [this approach](https://webapi.arcus-azure.net/features/security/auth/certificate#Globally-enforce-certificate-authentication) where it's a two-step but prefer to use a more user-friendly approach.

If we agree we can align our other filters.

🚧 This is WIP and will not be merged 🚧

### DIFF
--- a/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
+++ b/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
@@ -6,7 +6,7 @@ namespace Arcus.WebApi.Security.Authorization
     {
         private const string DefaultHeaderName = "x-managed-identity-token";
 
-        public string HeaderName { get; }
+        public string HeaderName { get; set; }
 
         public AzureManagedIdentityAuthorizationOptions(string headerName)
         {
--- a/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
+++ b/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
@@ -1,0 +1,20 @@
+﻿using GuardNet;
+
+namespace Arcus.WebApi.Security.Authorization
+{
+    public class AzureManagedIdentityAuthorizationOptions
+    {
+        private const string DefaultHeaderName = "x-managed-identity-token";
+
+        public string HeaderName { get; }
+
+        public AzureManagedIdentityAuthorizationOptions(string headerName)
+        {
+            Guard.NotNullOrEmpty(headerName,nameof(headerName));
+
+            HeaderName = headerName;
+        }
+
+        public static AzureManagedIdentityAuthorizationOptions Default = new AzureManagedIdentityAuthorizationOptions(DefaultHeaderName);
+    }
+}
--- a/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
+++ b/src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
@@ -15,6 +15,6 @@ namespace Arcus.WebApi.Security.Authorization
             HeaderName = headerName;
         }
 
-        public static AzureManagedIdentityAuthorizationOptions Default = new AzureManagedIdentityAuthorizationOptions(DefaultHeaderName);
+        public static readonly AzureManagedIdentityAuthorizationOptions Default = new AzureManagedIdentityAuthorizationOptions(DefaultHeaderName);
     }
 }
--- a/src/Arcus.WebApi.Security.Authorization/Extensions/FilterCollectionExtensions.cs
+++ b/src/Arcus.WebApi.Security.Authorization/Extensions/FilterCollectionExtensions.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using Arcus.WebApi.Security.Authorization;
+using GuardNet;
+using Microsoft.Extensions.DependencyInjection;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.AspNetCore.Mvc.Filters
+{
+    public static class FilterCollectionExtensions
+    {
+        /// <summary>
+        /// Adds Azure Managed Identity authorization
+        /// </summary>
+        /// <param name="filters">All filters that are being applied to the request pipeline</param>
+        /// <param name="serviceDependencies">Dependencies in the current application</param>
+        public static FilterCollection AddAzureManagedIdentityAuthorization(this FilterCollection filters, IServiceCollection serviceDependencies)
+        {
+            Guard.NotNull(filters, nameof(filters));
+
+            return filters.AddAzureManagedIdentityAuthorization(serviceDependencies, (options) => options);
+        }
+
+        /// <summary>
+        /// Adds Azure Managed Identity authorization
+        /// </summary>
+        /// <param name="filters">All filters that are being applied to the request pipeline</param>
+        /// <param name="serviceDependencies">Dependencies in the current application</param>
+        /// <param name="options">Configuration options for using Azure Managed Identity authorization</param>
+        public static FilterCollection AddAzureManagedIdentityAuthorization(this FilterCollection filters, IServiceCollection serviceDependencies, Func<AzureManagedIdentityAuthorizationOptions, AzureManagedIdentityAuthorizationOptions> options)
+        {
+            Guard.NotNull(filters, nameof(filters));
+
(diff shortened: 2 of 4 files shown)


src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using GuardNet;
+
+namespace Arcus.WebApi.Security.Authorization
+{
+    public class AzureManagedIdentityAuthorizationOptions
+    {
+        private const string DefaultHeaderName = "x-managed-identity-token";
+
+        public string HeaderName { get; set; }
```
fgheysels, line 9: I'd make this a readonly property.

fgheysels, line 9:
```suggestion
        public string HeaderName { get; }
```

src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        public AzureManagedIdentityAuthorizationOptions(string headerName)
+        {
+            Guard.NotNullOrEmpty(headerName,nameof(headerName));
+
+            HeaderName = headerName;
+        }
+
+        public static AzureManagedIdentityAuthorizationOptions Default = new AzureManagedIdentityAuthorizationOptions(DefaultHeaderName);
```
fgheysels, line 18:
```suggestion
        public static readonly AzureManagedIdentityAuthorizationOptions Default = new AzureManagedIdentityAuthorizationOptions(DefaultHeaderName);
```

src/Arcus.WebApi.Security.Authorization/Extensions/FilterCollectionExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <param name="serviceDependencies">Dependencies in the current application</param>
+        /// <param name="options">Configuration options for using Azure Managed Identity authorization</param>
+        public static FilterCollection AddAzureManagedIdentityAuthorization(this FilterCollection filters, IServiceCollection serviceDependencies, Action<AzureManagedIdentityAuthorizationOptions> options)
+        {
+            Guard.NotNull(filters, nameof(filters));
+
+            var azureManagedIdentityAuthorizationOptions = AzureManagedIdentityAuthorizationOptions.Default;
+            options(azureManagedIdentityAuthorizationOptions);
+
+            serviceDependencies.AddTransient(services => azureManagedIdentityAuthorizationOptions);
```
fgheysels, line 36: Should this be a transient scoped ?  Can possibly be a singleton

src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
tomkerkhove: If you do that, people won't be able to change the default value to whatevery they want to use

fgheysels: Hmm, weird.  I can commit my own suggestions :)

src/Arcus.WebApi.Security.Authorization/Extensions/FilterCollectionExtensions.cs
tomkerkhove: This can be for sure, I tend to not use singletons except for special reasons but I'm not an expert.

Happy to change if you want.

stijnmoreels:
@tomkerkhove I'll wait until it's out of draft-state to review?

src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
fgheysels: I'd say you'd do that (chaning the headername) by using the constructor ?

tomkerkhove:
It's best to review now because it will always be in draft since it's a design review.

Goal is to have a few pair of eyes look at the approach that is taken before it gets full implemented.

For example:
- Do we use the fluent approach for filters, but require `IServiceCollection`, or do we split it and expect consumers to call both methods?

src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
fgheysels: Ah, ok now I see.
You want it to behave like this:
```csharp
.AddAuthorizationFilter( o => o.HeaderName = "foo");
```
Ok then...

tomkerkhove: So you'd rather want to create the object yourself rather than a fluent model?

See https://github.com/arcus-azure/arcus.webapi/pull/147/files#diff-3399f8d754d976ba2fdf249b575dff02R91 for example

src/Arcus.WebApi.Security.Authorization/Extensions/FilterCollectionExtensions.cs
fgheysels: I'd say we can always use the same instance and avoid having new instances be created each time an instance is required.
What is your opinion @stijnmoreels ?

stijnmoreels: I think you should use the `serviceDependencies.Configure` for options. And use `IOptions<>` in the filter.

src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
stijnmoreels: Yes, I think I've the same idea as @fgheysels to use the object here since it's more clear what filter you gonna add + it's the current appraoch on all filters in the project.

src/Arcus.WebApi.Security.Authorization/Extensions/FilterCollectionExtensions.cs
tomkerkhove: I can do that!

src/Arcus.WebApi.Security.Authorization/AzureManagedIdentityAuthorizationOptions.cs
tomkerkhove: > Yes, I think I've the same idea as @fgheysels to use the object here since it's more clear what filter you gonna add + it's the current appraoch on all filters in the project.

Can you elaborate on that? Not sure if I'm following.

You're saying we should always go over ctor? If so, what do you propose for a consumer friendly appraoch of configuring it?

stijnmoreels: Bc we're currently using that type-approach, I'd say yes, favoring the ctor, but that can be injected. We can pass along the time and call the `.Configure` on the service collection later. (Like the .NET startup is mostly set up I guess).

stijnmoreels: But we can of course provide such an extension, it just seems more in line with the other built-in components that you 'register' the type somewhere and 'configure' it somewhere else.

tomkerkhove: While that's true, I'm wondering if that is the approach we want to use.

I'd rather have our users do: `filter.AddXXX` instead of having to do two calls, which they might miss in the docs and have to provide support on it.

But it might be just me.

stijnmoreels: Yes, I know. It's a bit discovery against user-friendly. Too much extensions make it not discoverable. While two or sometimes even more is not very user-friendly.
Can we provide both? Make the options and filter public enough to be used in both contexts and provide and extension that do both things?

fgheysels: Although I very much like immutable types, I think we should go with the way that is 'generally accepted' in asp.net, and that is the initial proposal of Tom:

Something like this:

```csharp
.AddAuthorizationFilter(options => options.HeaderName = "myFooBarHeader");
```

PR #148, opened 2020-04-22
Testing - add tests for Azure Managed Identity authorization filter

stijnmoreels:
Relates to #120

### DIFF
--- a/src/Arcus.WebApi.Security/Authorization/AzureManagedIdentityAuthorizationFilter.cs
+++ b/src/Arcus.WebApi.Security/Authorization/AzureManagedIdentityAuthorizationFilter.cs
@@ -8,7 +8,7 @@ namespace Arcus.WebApi.Security.Authorization
 {
     public class AzureManagedIdentityAuthorizationFilter : IAsyncAuthorizationFilter
     {
-        private const string DefaultHeaderName = "x-managed-identity-token";
+        public const string DefaultHeaderName = "x-managed-identity-token";
 
         private readonly string _headerName;
         private readonly IJwtTokenReader _jwtTokenReader;
--- a/src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
+++ b/src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
@@ -55,7 +55,7 @@ namespace Arcus.WebApi.Security.Authorization.Jwt
         }
 
         /// <summary>
-        ///     Validates if the token is considered valid.
+        ///     Verify if the token is considered valid.
         /// </summary>
         /// <param name="token">JWT token</param>
         public async Task<bool> IsValidTokenAsync(string token)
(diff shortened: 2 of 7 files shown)


gverstraete:
As promised, some feedback today. I think if I look at this, this looks fine, but to me this is implemented more as a generic and not specifically to Managed Identity. It seems more like a general token validator. So I would name it like that as well. Apart from that, nice work!

src/Arcus.WebApi.Security/Authorization/AzureManagedIdentityAuthorizationFilter.cs
```diff
@@ -8,7 +8,7 @@ namespace Arcus.WebApi.Security.Authorization
 {
     public class AzureManagedIdentityAuthorizationFilter : IAsyncAuthorizationFilter
```
stijnmoreels, line 9: The filter and reader should also be reviewed for XML docs. But obviously, not in this PR immediatly.

tomkerkhove:
@gverstraete So you're good to go with this?

With regards to the name, I was thinking the same recently so we might want to rename it for sure (too bad we actually shipped it already so that's another deprecation 😂)

Do you agree @fgheysels @krottiers

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -55,33 +55,44 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
         }
 
         /// <summary>
-        ///     Validates if the token is considered valid
+        ///     Validates if the token is considered valid.
```
fgheysels, line 58: I'd say "Verify if the token is valid" instead :)

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -55,33 +55,44 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
-
-            SecurityToken jwtToken;
             try
             {
-                _handler.ValidateToken(token, validationParameters, out jwtToken);
+                TokenValidationParameters validationParameters = await DetermineTokenValidationParametersAsync();
+                _handler.ValidateToken(token, validationParameters, out SecurityToken jwtToken);
             }
-            catch (Exception)
+            catch (Exception exception)
```
fgheysels, line 68: ~~What might be the reason of an exception that is thrown here ?
Do you want to 'hide' all kind of exceptions and then just return 'the token is not valid', and hide the actual cause of the exception ?~~

~~I'd say that you should catch specific exceptions: the ones that you know that can occur, and act accordingly.  For other exceptions, I think I would just let them bubble up the stack.~~

edit: I've been looking at the [documentation ](JwtSecurityTokenHandler )of that method and now I can understand why you catch on `Exception` :)

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
stijnmoreels: Done that.

stijnmoreels:
> @gverstraete So you're good to go with this?
> 
> With regards to the name, I was thinking the same recently so we might want to rename it for sure (too bad we actually shipped it already so that's another deprecation 😂)
> 
> Do you agree @fgheysels @krottiers

Added https://github.com/arcus-azure/arcus.webapi/issues/149

PR #156, opened 2020-05-06
Use configuration approach for JWT Token authorization

tomkerkhove:
Use configuration approach for JWT Token authorization.

Relates to #120

### DIFF
--- a/docs/preview/features/security/auth/jwt.md
+++ b/docs/preview/features/security/auth/jwt.md
@@ -1,0 +1,65 @@
+---
+title: "Authorization with JWT via ASP.NET Core authorization filters"
+layout: default
+---
+
+# Authorization with JWT
+
+The `Arcus.WebApi.Security` package provides a mechanism that uses JWT (JSON Web Tokens) to authorize requests access to the web application.
+
+This authorization process consists of the following parts:
+1. Find the OpenID server endpoint to request the correct access token
+2. Determine the request header name you want to use where the access token should be available
+
+## Installation
+
+This feature requires to install our NuGet package:
+
+```shell
+PM > Install-Package Arcus.WebApi.Security
+```
+
+## Usage
+
+The `JwtTokenAuthorizationFilter` can be added to the request filters in an <span>ASP.NET</span> Core application.
+This filter will then add authorization to all endpoints via the configured properties on the filter itself.
+
+Example:
+
+```csharp
+public class Startup
+{
+    public void ConfigureServices(IServiceCollection services)
+    {
+        // Default configuration:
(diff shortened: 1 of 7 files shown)


tomkerkhove:
Since I've been a blocker for this for a while I suggest you maybe pickup the tests & docs @stijnmoreels?

I don't have time soon to proceed with this and is only remaining thing for 1.0

stijnmoreels:
Yes, okay.

src/Arcus.WebApi.Security/Authorization/Extensions/FilterCollectionExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <param name="configureOptions">Configuration options for using JWT token authorization</param>
+        public static FilterCollection AddJwtTokenAuthorization(
+            this FilterCollection filters, 
+            Action<JwtTokenAuthorizationOptions> configureOptions)
+        {
+            Guard.NotNull(filters, nameof(filters));
+
+            var options = new JwtTokenAuthorizationOptions();
+            configureOptions?.Invoke(options);
+            filters.Add(new JwtTokenAuthorizationFilter(options));
```
stijnmoreels, line 34: Changed this bc one cannot use the options approach this late in the pipeline, this had to be configured before this, and since we are not a fan of that approach, we should use it directly like this.

docs/preview/features/security/auth/jwt.md
```diff
@@ -0,0 +1,65 @@
...
+            mvcOptions.Filters.AddJwtTokenAuthorization(options => options.JwtTokenReader = new JwtTokenReader(applicationId));
+        });
+
+        // Custom OpenID endpoint:
+        // You can use your own custom OpenID endpoint by providing another the endpoint in the options; additionally with custom validation parameters how the JWT token should be validated.
+        services.AddMvc(mvcOptions => 
+        {
+            var parameters = new TokenValidationParameters();
+            string endpoint = "https://localhost:5000/.well-known/openid-configuration";
+            mvcOptions.Filters.AddJwtTokenAuthorization(options => options.JwtTokenReader = new JwtTokenReader(parameters, endpoint));
```
stijnmoreels, line 61: One can argue that we could provide additional consumer-friendly methods like:
```csharp
options => options.WithCustomEndpoint(endpoint)
options => options.WithMicrosoftOnline(applicationId)
...
```

tomkerkhove: That can be a nice addition but not a must for now imo

src/Arcus.WebApi.Security/Authorization/Extensions/FilterCollectionExtensions.cs
tomkerkhove: Makes sense, thanks for fixing!

PR #168, opened 2020-07-09
Feature - update observability & security packages

stijnmoreels:
We can update the Arcus dependencies here bc it's not breaking changes, right?


tomkerkhove:
Sounds fair

PR #171, opened 2020-07-13
Feature - use HTTP correlation in Azure Functions

### LINKED ISSUE #139: Make it easy to use correlation with Azure Function
Make it easy to use correlation with Azure Function to use correlation for example.

stijnmoreels:
Closes #139


docs/preview/features/correlation.md
```diff
@@ -108,4 +108,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+
+When this addition is added, you can use the `HttpCorrelationService` inside your function to call the correlation functionality and use the `ICorrelationInfoAccessor` (like before) to have access to the `CorrelationInfo` of the HTTP request.
+
+```csharp
+public class MyHttpFunction
+{
+    private readonly HttpCorrelationService _correlationService;
+    private readonly ICorrelationInfoAccessor _correlationAccessor;
+
+    public MyHttpFunction(HttpCorrelationService correlationService, ICorrelationInfoAccessor correlationAccessor)
```
stijnmoreels, line 148: Moved the HTTP correlation functionality to a service, used in both Azure Functions and API projects. But not sure if we should merge this with the correlation accessor in the Azure Functions variant...
There are also projects that create some kind of middleware alternative for Azure Functions but that's maybe an idea if we have additional functionality we want to move to Azure Functions.

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationInfoAccessor.cs
```diff
@@ -38,7 +39,7 @@ public CorrelationInfo GetCorrelationInfo()
         public void SetCorrelationInfo(CorrelationInfo correlationInfo)
```
fgheysels, line 39: Maybe implement this method explicitly ?

stijnmoreels: To hide it you mean? That wouldn't do much bc we use the interfaces in the application. In fact, we could maybe make this class `internal` bc the consumer doesn't interact with it this way.

src/Arcus.WebApi.Logging/Correlation/CorrelationMiddleware.cs
```diff
@@ -50,94 +40,15 @@ public async Task Invoke(HttpContext httpContext)
...
             {
-                string operationId = _options.Operation.GenerateId();
-                if (String.IsNullOrWhiteSpace(operationId))
-                {
-                    throw new InvalidOperationException(
-                        $"Correlation cannot use '{nameof(_options.Operation.GenerateId)}' to generate an operation ID because the resulting ID value is blank");
-                }
-
-                return operationId;
+                httpContext.Response.StatusCode = StatusCodes.Status400BadRequest;
```
tomkerkhove, line 49: This is a behavioral change which should not be the case I think. Is there a reason why this was changed? When our correlation fails I don't expect it to give a bad request back?

docs/preview/features/correlation.md
```diff
@@ -108,4 +108,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+        _correlationService = correlationService;
+        _correlationAccessor = correlationAccessor;
+    }
+
+    [FunctionName("Function1")]
+    public async Task<IActionResult> Run(
+        [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
+        ILogger log)
+    {
+        if (_correlationService.TryHttpCorrelate(out string errorMessage))
```
tomkerkhove, line 159: Did you successfully run this in an Azure Function? 🤔 

Would be amazed if it works, if so - Even better!

docs/preview/features/correlation.md
tomkerkhove: What if we just call it `HttpCorrelation`?

In terms of the alternative, do you have a link to an example?

docs/preview/features/correlation.md
```diff
@@ -108,4 +108,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+    private readonly HttpCorrelationService _correlationService;
+    private readonly ICorrelationInfoAccessor _correlationAccessor;
+
+    public MyHttpFunction(HttpCorrelationService correlationService, ICorrelationInfoAccessor correlationAccessor)
+    {
+        _correlationService = correlationService;
+        _correlationAccessor = correlationAccessor;
+    }
+
+    [FunctionName("Function1")]
```
tomkerkhove, line 154:
```suggestion
    [FunctionName("HTTP-Correlation-Example")]
```

docs/preview/features/correlation.md
```diff
@@ -108,4 +108,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+```csharp
+public class MyHttpFunction
+{
+    private readonly HttpCorrelationService _correlationService;
+    private readonly ICorrelationInfoAccessor _correlationAccessor;
+
+    public MyHttpFunction(HttpCorrelationService correlationService, ICorrelationInfoAccessor correlationAccessor)
+    {
+        _correlationService = correlationService;
+        _correlationAccessor = correlationAccessor;
```
tomkerkhove, line 151: What if you could access the correlation accessor through the HttpCorrelation object?

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationInfoAccessor.cs
tomkerkhove: But it can't hurt to implement it and register the feature on HTTP Context again, if it's not much work?

tomkerkhove: If folks want to use it then, they can

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationService.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            AddCorrelationResponseHeaders(httpContext, operationId, transactionId);
+
+            errorMessage = null;
+            return true;
+        }
+
+        private string DetermineOperationId(HttpContext httpContext)
+        {
+            if (String.IsNullOrWhiteSpace(httpContext.TraceIdentifier))
```
tomkerkhove, line 87: And this is the fun part of Azure Functions, they do not use this approach (unless this changed); see https://github.com/Azure/azure-functions-host/issues/5014

That's why we are using this for Azure Functions:
```csharp
private string GetOperationIdFromRequestHeaders(ExecutionContext executionContext)
{
    return Activity.Current?.Id ?? executionContext.InvocationId.ToString();
}
```

tomkerkhove: So we can use this for sure, but you'll need to use different operation id providers depending on the runtime

docs/preview/features/correlation.md
stijnmoreels: > In terms of the alternative, do you have a link to an example?

https://github.com/umamimolecule/azure-functions-http-middleware#introduction

stijnmoreels: Suggested renaming is good.

src/Arcus.WebApi.Logging/Correlation/CorrelationMiddleware.cs
stijnmoreels: Ah, no, it's not actually a behavior change. It's something that's being moved. It was first being set direclty in the middleware but since we now have extracted the common logic, we haven't got access to certain models. (like in the azure functions you should create an `IActionResult` as respnose but not in the middleware).

docs/preview/features/correlation.md
stijnmoreels: Yes, I ran several manual tests.

docs/preview/features/correlation.md
stijnmoreels: Yes, that was something that I also suggested [here](https://github.com/arcus-azure/arcus.webapi/pull/171#discussion_r453570179). It could be added value since you then only have to inject 1 correlation model.

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationService.cs
stijnmoreels: Hmm, I think I got a valid reponse back with the `.TraceIdentifier` filled out. 🤔 .

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationInfoAccessor.cs
stijnmoreels: Ok, implemented and tested.

docs/preview/features/correlation.md
tomkerkhove: Even better!

tomkerkhove: Might be a good integration test to add in the future maybe?

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationService.cs
tomkerkhove: Do you have the function in a repo somewhere?

stijnmoreels: No, but I'll look if I can maybe create an integration test for this; with an embedded Azure Function.

stijnmoreels: No, not that. Hmm, am thinking what would be the best approach to test this.

tomkerkhove: I would add an Azure Function to the solution which we start in the pipeline and send request to, I would not start if from our code.

tomkerkhove: You can use the same approach locally.

src/Arcus.WebApi.Tests.Projects.AzureFunction/HttpTriggerFunction.cs
```diff
@@ -0,0 +1,43 @@
...
+using Arcus.Observability.Correlation;
+using Arcus.WebApi.Logging.Correlation;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Arcus.WebApi.Tests.Projects.AzureFunction
```
tomkerkhove, line 13: Nit - Maybe this is better name?
```suggestion
namespace Arcus.WebApi.Tests.Runtimes.AzureFunction
```

src/Arcus.WebApi.Tests.Projects.AzureFunction/HttpTriggerFunction.cs
```diff
@@ -0,0 +1,43 @@
...
+        public async Task<IActionResult> Run(
+            [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
+            ILogger log)
+        {
+            if (_correlationService.TryHttpCorrelate(out string errorMessage))
+            {
+                return new BadRequestObjectResult(errorMessage);
+            }
+
+            string json = JsonConvert.SerializeObject(_correlationInfoAccessor.GetCorrelationInfo());
```
tomkerkhove, line 39: Maybe log the information before deserializing for sake of troubleshooting?

src/Arcus.WebApi.Tests.Integration/Logging/Correlation/AzureFunctionCorrelationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Arcus.WebApi.Tests.Integration.Logging.Correlation
+{
+    public class AzureFunctionCorrelationTests
+    {
+        private const string DefaultOperationId = "RequestId",
+                             DefaultTransactionId = "X-Transaction-ID",
+                            DefaultRoute = "http://localhost:5000/";
```
tomkerkhove: Let's parameterize the port since it's duplicated in the pipeline as well

build/ci-build.yml
```diff
@@ -89,6 +91,18 @@ stages:
...
+              dockerFile: ./src/$(ProjectName)/Dockerfile
+              imageName: '$(ProjectName):$(Build.BuildId)'
+          - task: Docker@1
+            displayName: 'Run new  project Docker image'
+            inputs:
+              command: 'Run an image'
+              imageName: '$(ProjectName):$(Build.BuildId)'
+              containerName: '$(ProjectName)'
+              ports: '5000'
           - template: test/run-integration-tests.yml@templates
```
tomkerkhove, line 112: I've found it usefule to run this step after running the tests: https://github.com/tomkerkhove/promitor/blob/master/build/azure-devops/agents-ci-discovery.yml#L141

Note that it always runs,even if the tests fail. That way you can use the logs during troubleshooting.

src/Arcus.WebApi.Tests.Integration/Logging/Correlation/AzureFunctionCorrelationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+                Assert.False(String.IsNullOrWhiteSpace(content.TransactionId), "Accessed 'X-Transaction-ID' cannot be blank");
+                Assert.False(String.IsNullOrWhiteSpace(content.OperationId), "Accessed 'X-Operation-ID' cannot be blank");
+
+                Assert.Equal(correlationId, content.TransactionId);
+                Assert.Equal(requestId, content.OperationId);
+            }
+        }
+
+        [Fact]
+        public async Task SendRequest_WithCorrelationHeader_ResponseWithSameCorrelationHeader()
```
tomkerkhove, line 42:
```suggestion
        public async Task SendRequest_WithTransactionIdHeader_ResponseWithSameTransactionIdHeader()
```

docs/preview/features/correlation.md
tomkerkhove: Let's do it!

docs/preview/features/correlation.md
tomkerkhove: Thanks for adding!

src/Arcus.WebApi.Tests.Projects.AzureFunction/HttpTriggerFunction.cs
stijnmoreels: Hmm, I don't find that more clearly. 'Runtimes' makes me think of not 'complete' projects while in fact it is a fully functional project used in the test suite.

docs/preview/features/correlation.md
stijnmoreels: I'll added it as a property, but it got me wondering if we should let the `HttpCorrelation` implement the `ICorrelationInfoAccessor` interface and delegate it to (now private) field.

src/Arcus.WebApi.Tests.Projects.AzureFunction/HttpTriggerFunction.cs
tomkerkhove: It feels to me since we are running tests against the Functions runtime but it's ok. Project doesn't say a thing if you ask me if you open up the solution, but that's just my opinion

build/ci-build.yml
stijnmoreels: Added! Thanks.

docs/preview/features/correlation.md
tomkerkhove: I would let `HttpCorrelation` implement `ICorrelationInfoAccessor` so that I can use that as my single place for HTTP correlation.


> it got me wondering if we should let the `HttpCorrelation` implement the `ICorrelationInfoAccessor` interface and delegate it to (now private) field.

If you are talking about delegating the fields, why not simply use `HttpCorrelationInfoAccessor` inside `HttpCorrelation` instead of implementing it or am I misunderstanding your suggestion?

stijnmoreels:
Hmm, not sure anymore what I've done wrong 🤔 . I thought if you run an Docker application with `5000:80`, it will use `5000` public and `80` private? Bc I haven't found a way to change the port number in Azure Functions, I thought I could fix it like this. But apparently, this gives `Unauthorized` for some reason. 🤔 I'll investigate further.

docs/preview/features/correlation.md
stijnmoreels: > I would let HttpCorrelation implement ICorrelationInfoAccessor so that I can use that as my single place for HTTP correlation.

That's actually my suggestion. 😄 . So, yes, some misunderstanding here.

src/Arcus.WebApi.Tests.Projects.AzureFunction/HttpTriggerFunction.cs
stijnmoreels: Ok, I'll change it.

stijnmoreels:
> Hmm, not sure anymore what I've done wrong 🤔 . I thought if you run an Docker application with `5000:80`, it will use `5000` public and `80` private? Bc I haven't found a way to change the port number in Azure Functions, I thought I could fix it like this. But apparently, this gives `Unauthorized` for some reason. 🤔 I'll investigate further.

Ok, I found every issue now and fixed it.

build/nuget-release.yml
```diff
@@ -18,6 +18,7 @@ variables:
   - group: 'Arcus - GitHub Package Registry'
   - group: 'Build Configuration'
   - template: ./variables/build.yml
+  - template: ./variables/test.yml
```
tomkerkhove, line 21: You'll need the function to be running in this pipeline as well

src/Arcus.WebApi.Tests.Runtimes.AzureFunction/HttpTriggerFunction.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+        
+        [FunctionName("HttpTriggerFunction")]
+        public async Task<IActionResult> Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequest req,
+            ILogger log)
+        {
+            if (_correlationService.TryHttpCorrelate(out string errorMessage))
+            {
+                CorrelationInfo correlationInfo = _correlationService.CorrelationInfoAccessor.GetCorrelationInfo();
```
tomkerkhove, line 34: Ok now I get your comment, here I'd expect this to be possible:
```suggestion
                CorrelationInfo correlationInfo = _correlationService.GetCorrelationInfo();
```

Which just calls `CorrelationInfoAccessor` under the hood for me.

docs/preview/features/correlation.md
tomkerkhove: I like the current approach, just added a remark in the test file; if they still want to use them separately they can inject both; but you're not forced to.

docs/preview/features/correlation.md
```diff
@@ -108,4 +108,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 }
 ```
 
+## Using secret store within Azure Functions
```
tomkerkhove, line 111: Let's update the docs according to how it works now.

src/Arcus.WebApi.Tests.Runtimes.AzureFunction/HttpTriggerFunction.cs
stijnmoreels: Hmm, no that was actually my recent consideration, that moving the property would maybe make it more 'weird' as we're mixing concepts. I think here with the property makes the 'accessing' and 'determination' of correlation separate.

docs/preview/features/correlation.md
stijnmoreels: Updated the suggessted approach and added a note to how you can use your own injected `ICorrelationInfoAccessor`.

build/nuget-release.yml
stijnmoreels: Moved to template and added reference here.

src/Arcus.WebApi.Tests.Runtimes.AzureFunction/HttpTriggerFunction.cs
tomkerkhove: I see it more as an extension and it uses the accessor but it provides more features; but the consumer shouldn't have to care/know that it's using the accessor under the hood imo

docs/preview/features/correlation.md
```diff
@@ -108,4 +108,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+```
+
+When this addition is added, you can use the `HttpCorrelation` inside your function to call the correlation functionality and use the `ICorrelationInfoAccessor` (like before) to have access to the `CorrelationInfo` of the HTTP request.
+
+```csharp
+public class MyHttpFunction
+{
+    private readonly HttpCorrelation _correlationService;
+
+    public MyHttpFunction(HttpCorrelation correlationService)
```
tomkerkhove, line 147: maybe rename `correlationService?`

docs/preview/features/correlation.md
```diff
@@ -108,4 +108,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+    [FunctionName("HTTP-Correlation-Example")]
+    public async Task<IActionResult> Run(
+        [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
+        ILogger log)
+    {
+        if (_correlationService.TryHttpCorrelate(out string errorMessage))
+        {
+            log.LogInformation("C# HTTP trigger function processed a request.");
+
+            CorrelationInfo correlationInfo = _correlationService.CorrelationInfoAccessor.GetCorrelationInfo();
```
tomkerkhove, line 161:
```suggestion
            // Easily access correlation information in your application
            CorrelationInfo correlationInfo = _correlationService.CorrelationInfoAccessor.GetCorrelationInfo();
```

tomkerkhove:
Once the open conversations are done we are good to go I think! Sorry for taking so long..

src/Arcus.WebApi.Tests.Runtimes.AzureFunction/HttpTriggerFunction.cs
stijnmoreels: OK, moved.

docs/preview/features/correlation.md
stijnmoreels: Yes, renamed to `CorrelationService correlationService`.

tomkerkhove: Actually I just meant the variable name, not the type name as HttpCorrelation is what it's doing if you ask me.

I would revert the renaming from https://github.com/arcus-azure/arcus.webapi/pull/171/commits/3e3a7011d206cf2658aba3f263fb197c3cb22f16

stijnmoreels: 😅 Yeah, I already thought there was something off.

tomkerkhove: Great, but can you still rename the variable please? :D

PR #174, opened 2020-08-15
Don't dispose stream when logging request bodies

### LINKED ISSUE #165: UseRequestTracking  with IncludeBodyRequest throws an exception
I've an API in which I have RequestTracking enabled by using these lines of code:

```csharp
app.UseRequestTracking(options => options.IncludeRequestBody = true);
app.UseHttpCorrelation();
```

As you can see, I've enabled RequestTracking and I want to have the body of the requests logged to Application Insights for instance.

When executing the API operation, the API operation returns successfully and the response is correctly being returned.
However, when I look into Application Insights, I see this:

image

When clicking on an item, it seems that something wants to read a closed stream:

>Cannot access a closed Stream.

stacktrace:
```
System.ObjectDisposedException:
   at System.IO.MemoryStream.get_Position (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at Microsoft.AspNetCore.WebUtilities.FileBufferingReadStream.get_Position (Microsoft.AspNetCore.WebUtilities, Version=3.1.3.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)
   at Arcus.WebApi.Logging.RequestTrackingMiddleware+<SanitizeRequestBodyAsync>d__7.MoveNext (Arcus.WebApi.Logging, Version=1.0.1.0, Culture=neutral, PublicKeyToken=null)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at Arcus.WebApi.Logging.RequestTrackingMiddleware+<TrackRequestAsync>d__5.MoveNext (Arcus.WebApi.Logging, Version=1.0.1.0, Culture=neutral, PublicKeyToken=null)
```

I'm using Arcus.WebApi.Logging 1.01

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

Relates to #165

### DIFF
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -105,35 +105,24 @@ namespace Arcus.WebApi.Logging
             return requestHeaders.Where(header => _options.OmittedHeaderNames.Contains(header.Key) == false);
         }
 
-        private static async Task<string> SanitizeRequestBodyAsync(Stream requestStream)
+        private async Task<string> SanitizeRequestBodyAsync(Stream requestStream)
         {
             if (!requestStream.CanRead)
             {
+                _logger.LogWarning("Request body could not be tracked because stream is not readable");
                 return null;
             }
 
-            long originalPosition = 0;
-            if (requestStream.CanSeek)
+            if (!requestStream.CanSeek)
             {
-                if (requestStream.Position != 0)
-                {
-                    originalPosition = requestStream.Position;
-                    requestStream.Position = 0;
-                }
+                _logger.LogWarning("Request body could not be tracked because stream is not seekable");
+                return null;
             }
-            else
+
+            long originalPosition = requestStream.Position;
+            if (requestStream.Position != 0)
             {
-                try
-                {
-                    if (requestStream.Position != 0)
-                    {
-                        return null;
-                    }
-                }
-                catch
-                {
-                    return null;
-                }
+                requestStream.Seek(0, SeekOrigin.Begin);
             }
 
             var reader = new StreamReader(requestStream);
@@ -141,10 +130,7 @@ namespace Arcus.WebApi.Logging
 
             try
             {
-                if (requestStream.CanSeek)
-                {
-                    requestStream.Seek(originalPosition, SeekOrigin.Begin);
-                }
+                requestStream.Seek(originalPosition, SeekOrigin.Begin);
             }
             catch
             {
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -87,7 +87,7 @@ namespace Arcus.WebApi.Logging
                     }
                 }
 
-                Dictionary<string, object> logContext = telemetryContext.ToDictionary(kv => kv.Key, kv => (object) kv.Value);
+                Dictionary<string, object> logContext = telemetryContext.ToDictionary(kv => kv.Key, kv => (object)kv.Value);
                 _logger.LogRequest(httpContext.Request, httpContext.Response, duration, logContext);
             }
             catch (Exception exception)
@@ -112,10 +112,12 @@ namespace Arcus.WebApi.Logging
                 return null;
             }
 
+            long originalPosition = 0;
             if (requestStream.CanSeek)
             {
                 if (requestStream.Position != 0)
                 {
+                    originalPosition = requestStream.Position;
                     requestStream.Position = 0;
                 }
             }
@@ -134,11 +136,22 @@ namespace Arcus.WebApi.Logging
                 }
             }
 
-            using (var reader = new StreamReader(requestStream))
+            var reader = new StreamReader(requestStream);
+            string contents = await reader.ReadToEndAsync();
+
+            try
             {
-                string contents = await reader.ReadToEndAsync();
-                return contents;
+                if (requestStream.CanSeek)
+                {
+                    requestStream.Seek(originalPosition, SeekOrigin.Begin);
+                }
             }
+            catch
+            {
+                // Nothing to do here, we want to ensure the value is always returned.
+            }
+
+            return contents;
         }
     }
 }
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -7,7 +7,6 @@ using System.Threading.Tasks;
 using GuardNet;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Primitives;
 
 namespace Arcus.WebApi.Logging
@@ -109,14 +108,12 @@ namespace Arcus.WebApi.Logging
         {
             if (!requestStream.CanRead)
             {
-                _logger.LogWarning("Request body could not be tracked because stream is not readable");
-                return null;
+                return "Request body could not be tracked because stream is not readable";
             }
 
             if (!requestStream.CanSeek)
             {
-                _logger.LogWarning("Request body could not be tracked because stream is not seekable");
-                return null;
+                return "Request body could not be tracked because stream is not seekable";
             }
 
             long originalPosition = requestStream.Position;


tomkerkhove:
This should be it imo. What do you think @fgheysels?

Unfortunately, I'm not able to reproduce this in the tests and will see what I can come up with to simulate it.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -112,10 +112,12 @@ private static async Task<string> SanitizeRequestBodyAsync(Stream requestStream)
                 return null;
             }
 
+            long originalPosition = 0;
             if (requestStream.CanSeek)
             {
                 if (requestStream.Position != 0)
```
fgheysels, line 118: I don't know if I'd check if the current position is not 0, before setting the `originalPosition` and positioning the stream to the beginning.   But that's just my opinion

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -134,11 +136,22 @@ private static async Task<string> SanitizeRequestBodyAsync(Stream requestStream)
...
-                string contents = await reader.ReadToEndAsync();
-                return contents;
+                if (requestStream.CanSeek)
+                {
+                    requestStream.Seek(originalPosition, SeekOrigin.Begin);
+                }
             }
+            catch
+            {
+                // Nothing to do here, we want to ensure the value is always returned.
```
fgheysels, line 151: What exception would you expect here ?  And why not log the exception should one occur ?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: Good point!

I'll assign the original position regardless of the position

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: For me this is best effort. The app should not return 500 nor log exception (= alert) on this.

But I could live with a warning if you prefer.

fgheysels: Indeed, you don't want to have this logged for every request.
But then I was thinking; why not add a guard clause at the beginning of the method which verifies if the stream is seekable.  If not, return there already.   If the string is not seekable, you're going to read the request stream here, and the stream will be positioned at the end, making it unusable for later processing.

```
private static async Task<string> SanitizeRequestBodyAsync(Stream requestStream)
{	            
    if (!requestStream.CanRead)	            
    {	   
        return null;	           
    }	            

    if( !requestStream.CanSeek )
    {
        return null;
     }
```

tomkerkhove: What if it's in the middle of the stream (for some reason), do you still want to not look at it?

fgheysels: If the stream is not seekable, you're not able to reposition it to it's original position.   Do you want to read the stream then anyway and possibly cause side-effects ?

tomkerkhove: Good point, might align with you. Thoughts @stijnmoreels ?

tomkerkhove:
> Seems good to me, I think. Can we come up with a test with maybe a controller that reads the request's body (maybe not fully) so we can see that it is logged and reset correctly? Maybe with an extra middleware component that asserts.

Open for suggestions. What do you expect it to look like then?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: Yes, I think @fgheysels has a point. Shortcutting when not seekable and logging this?

stijnmoreels:
> > Seems good to me, I think. Can we come up with a test with maybe a controller that reads the request's body (maybe not fully) so we can see that it is logged and reset correctly? Maybe with an extra middleware component that asserts.
> 
> Open for suggestions. What do you expect it to look like then?

Prerequisites:
- Middleware with maybe logging the request's body position (sinks back to in-memory log sink that we can assert in our test)
- Controller that seeks the request stream (for example only half)

1. Test sends request (body 100%)
2. Goes through our request logging middleware (buffering enabled)
3. Test controller reads body parially (example 50%)
4. Goes through our request logging middleware
  - when body stream is seekable: should log entire body, reset to original position
  - when body stream is not seekable: should not log request body
5. Goes through our extra test middleware and logs the request body stream position
6. Asserts on
  - when body stream is seekable: gets entire original request body from in-memory log sink + expected request stream position (when 50% read, half of the stream)
  - when body stream is not seekable: doesn't get request body from in-memory log sink + expected request stream position (when 50% read, half of the stream)

stijnmoreels:
> > > Seems good to me, I think. Can we come up with a test with maybe a controller that reads the request's body (maybe not fully) so we can see that it is logged and reset correctly? Maybe with an extra middleware component that asserts.
> > 
> > 
> > Open for suggestions. What do you expect it to look like then?
> 
> Prerequisites:
> 
> * Middleware with maybe logging the request's body position (sinks back to in-memory log sink that we can assert in our test)
> * Controller that seeks the request stream (for example only half)
> 
> 1. Test sends request (body 100%)
> 2. Goes through our request logging middleware (buffering enabled)
> 3. Test controller reads body parially (example 50%)
> 4. Goes through our request logging middleware
>    
>    * when body stream is seekable: should log entire body, reset to original position
>    * when body stream is not seekable: should not log request body
> 5. Goes through our extra test middleware and logs the request body stream position
> 6. Asserts on
>    
>    * when body stream is seekable: gets entire original request body from in-memory log sink + expected request stream position (when 50% read, half of the stream)
>    * when body stream is not seekable: doesn't get request body from in-memory log sink + expected request stream position (when 50% read, half of the stream)


But maybe not in this PR; + we may need some input from @fgheysels to have somewhat the same setup so we can verify our fix solves this? But you had some problems, you said, reproducing?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: I've changed it but used warnings instead so that it's more explicit and an empty body is tracked.

Thoughts about this? Reason is that otherwise you will only notice if you check the tracked request and not have "indicators" of issues.

tomkerkhove:
I'll leave that up to you indeed @stijnmoreels, work your ✨

tomkerkhove:
> But maybe not in this PR; + we may need some input from @fgheysels to have somewhat the same setup so we can verify our fix solves this? But you had some problems, you said, reproducing?

Will see if I can reproduce it somehow when I'm back in the office next week as well.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
fgheysels: I'm just thinking on how I'd use it myself:

Suppose I have an API call that fails for some reason or another, I'd inspect the logs and one of the things that I might want to check is the request body.    If the body has been logged: good.  If it hasn't, and instead of the body being present in this property a string is present which says that the request body could not be logged because of this or that reason:  too bad, but at least I know why it hasn't been logged without having to inspect any other logs to find out why the body is not present in the logs.

I don't think that having this kind of information as a separate warning gives any additional value.  I'm also not going to search actively in my logs for requests for which the request body could not be logged I think :).

tomkerkhove: Makes sense!

stijnmoreels (approved):

PR #177, opened 2020-09-08
Feature - add trace logging during correlation & request tracking

stijnmoreels:
Provide extra trace logging when making decisions based on consumer-configured options. Explaining why the functionality behaves based on consumer-input so debugging and defect localization is more easily done.

### DIFF
--- a/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelation.cs
+++ b/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelation.cs
@@ -152,7 +152,7 @@ namespace Arcus.WebApi.Logging.Correlation
                     return newlyGeneratedTransactionId;
                 }
 
-                _logger.LogWarning(
+                _logger.LogTrace(
                     "No transactional correlation ID found in request header '{HeaderName}' but since the correlation options specifies that no transactional ID should be generated, there will be no ID present", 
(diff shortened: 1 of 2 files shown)


src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelation.cs
```diff
@@ -115,45 +115,59 @@ private string DetermineOperationId(HttpContext httpContext)
...
                         throw new InvalidOperationException(
                             $"Correlation cannot use function '{nameof(_options.Transaction.GenerateId)}' to generate an transaction ID because the resulting ID value is blank");
                     }
 
-                    return transactionId;
+                    _logger.LogTrace("Generated '{TransactionId}' as transactional correlation ID", newlyGeneratedTransactionId);
+                    return newlyGeneratedTransactionId;
                 }
 
+                _logger.LogWarning(
```
tomkerkhove, line 155: I'd go with a trace here as this can be noisy while they've turned it off as well

tomkerkhove (approved):

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelation.cs
stijnmoreels: Ok, changed. Resolving.

Issue #175, opened 2020-08-17
Allow optional jwt token validation using the oid


MichielVanwelsenaere:
**Is your feature request related to a problem? Please describe.**
In the current implementation the `JwtTokenAuthorizationFilter` verifies the `aud` value in the jwt token. That `aud` value is set to what is requested by the client and can be multiple things. I'm currently facing a scenario where we want to secure the last mile between APIM and backend App Services. The APIM instance is allowed to call the App Services and nothing else thus my idea was the following:
1. in the Apim policy request a MSI token (resource audience doesn't matter). Because the APIM resource is used to request the jwt token the `oid` will be automatically set to the MSI objectId of the APIM resource.
2. Validate in the app service code whether the jwt token passed has the expected `oid` value containing the MSI objectId of the APIM resource.

**Describe the solution you'd like**
An adjustment of the `JwtTokenAuthorizationFilter` so it allows for filtering on `oid` and/or `aud`. This depending on the needs of the application/developers. 

**Describe alternatives you've considered**
None 

**Additional context**
This suggested feature assumes the following:
1. it's not possible for any resource/client to request a jwt token with a configurable `oid` value on the Microsoft OpenID endpoint.

fgheysels:
Is `oid` something that is Microsoft specific ?  It seems to be the object-id of an Azure user / application ?  Wouldn't this be to 'Azure specific', or is that something that doesnt matter, as Arcus is already pretty tied to Azure ?

Can't we use `sub` instead ?  (Would need to investigate what it contains in this particular scenario)

MichielVanwelsenaere:
> 
> 
> Is `oid` something that is Microsoft specific ? It seems to be the object-id of an Azure user / application ? Wouldn't this be to 'Azure specific', or is that something that doesnt matter, as Arcus is already pretty tied to Azure ?
> 
> Can't we use `sub` instead ? (Would need to investigate what it contains in this particular scenario)

Could be that `oid` is Microsoft specific. `sub` has the same value as `oid` in this particular scenario.

MichielVanwelsenaere:
perhaps interesting as well could be `xms_mirid` as that contains the resourceId of the resource that requested the token. Perhaps it makes sense to allow checking of one or more configurable values (`aud`,`oid`, `xms_mirid`, etc..)?

gverstraete:
[...]

PR #179, opened 2020-09-14
Add behavior to allow custom claims validation (#175)

egeda9:
Following the discussion in [https://github.com/arcus-azure/arcus.webapi/issues/175#issuecomment-689984153](url) this proposal covers the constructor overload by using a **ClaimCheck** object and then the IsValidTokenAsync is able to identify if the class parameter contains any custom claim to verify it through comparison by name and value.

Relates to #175


src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -63,29 +80,52 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
             {
                 return false;
             }
 
-            return true;
+            return result;
+        }
+
+        /// <summary>
+        ///
```
tomkerkhove, line 100: Can you provide docs please?

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -63,29 +80,52 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="validatedToken"></param>
+        /// <returns></returns>
+        private bool ValidateClaimCheck(SecurityToken validatedToken)
+        {
+            if (!this._claimCheck.Any())
```
tomkerkhove, line 106: We tend to always add the {}

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -63,29 +80,52 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
+        /// <param name="validatedToken"></param>
+        /// <returns></returns>
+        private bool ValidateClaimCheck(SecurityToken validatedToken)
+        {
+            if (!this._claimCheck.Any()) 
+                return true;
+
+            var jwtToken = (JwtSecurityToken) validatedToken;
+
+            var validClaims = (from c in jwtToken.Claims
```
tomkerkhove, line 111: What's the goal here? I'm not sure I get the logic 🤔

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -53,8 +70,8 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
             Guard.NotNullOrWhitespace(openIdConnectDiscoveryUri, nameof(openIdConnectDiscoveryUri));
             Guard.NotNull(tokenValidationParameters, nameof(tokenValidationParameters));
 
-            _tokenValidationParameters = tokenValidationParameters;
-            _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(openIdConnectDiscoveryUri, new OpenIdConnectConfigurationRetriever());
+            this._tokenValidationParameters = tokenValidationParameters;
```
tomkerkhove, line 73: You can leave this out
```suggestion
            _tokenValidationParameters = tokenValidationParameters;
```

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -53,8 +70,8 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
+            this._configManager = new ConfigurationManager<OpenIdConnectConfiguration>(openIdConnectDiscoveryUri, new OpenIdConnectConfigurationRetriever());
```
tomkerkhove, line 74: You can leave this out
```suggestion
            _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(openIdConnectDiscoveryUri, new OpenIdConnectConfigurationRetriever());
```

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -17,10 +20,24 @@ public class JwtTokenReader : IJwtTokenReader
...
         private readonly TokenValidationParameters _tokenValidationParameters;
         private readonly JwtSecurityTokenHandler _handler = new JwtSecurityTokenHandler();
         private readonly ConfigurationManager<OpenIdConnectConfiguration> _configManager;
 
+        /// <summary>
+        ///     Constructor
+        /// </summary>
+        /// <remarks>Uses Microsoft OpenId connect discovery endpoint</remarks>
+        /// <param name="claimCheck">Custom claims key-value pair to validate against</param>
+        public JwtTokenReader(ClaimCheck claimCheck)
```
tomkerkhove, line 33: It's best to call `JwtTokenReader(TokenValidationParameters tokenValidationParameters, string openIdConnectDiscoveryUri)` overload and add claims there which is now a bit ignored

tomkerkhove, line 33: Do we need a custom type for this or can we just pass a dictionary? This feels a bit less overkill, no?

tomkerkhove (changes requested):
Did a first run, overall looks ok but needs some improvements.

Could you also:
- Provide documentation [here](https://github.com/arcus-azure/arcus.webapi/tree/master/docs/preview) please?
- Update how we register the authorization with an optional parameter for the claims : ([based on this](https://github.com/arcus-azure/arcus.webapi/issues/175#issuecomment-689984153))

```csharp
services.AddMvc(mvcOptions => 
{
    var parameters = new TokenValidationParameters();
    string endpoint = "https://localhost:5000/.well-known/openid-configuration";
    mvcOptions.Filters.AddJwtTokenAuthorization(claimChecks, options => options.JwtTokenReader = new JwtTokenReader(parameters, endpoint));
});
```

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
fgheysels, line 33: Is it sufficient to only allow one `ClaimCheck` or should we allow to have multiple ?
Maybe use a constructor that looks like this:

```csharp
public JwtTokenReader( params ClaimCheck[] claimChecks)
```

fgheysels: I want to leave a comment on line 45, but I'm unable to do so.

Should the constructor that takes the `applicationId` parameter call the new constructor ?  So in fact, the existing ctor can create a ClaimCheck instance for the `aud` claim, and use the `applicationId` as value ?

We can then also get rid of the `_applicationId` member.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -63,29 +80,52 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
             {
-                OpenIdConnectConfiguration config = await _configManager.GetConfigurationAsync();
-                
+                var config = await this._configManager.GetConfigurationAsync();
+
                 var validationParameters = new TokenValidationParameters
                 {
                     ValidateAudience = false,
-                    ValidAudience = _applicationId,
+                    ValidAudience = !this._claimCheck.Any() ? this._applicationId : string.Empty,
```
fgheysels, line 128: You can check if there exists a ClaimCheck for the audience, and use that one instead of this construct.
Next to that, why is `ValidateAudience` set to false if `ValidAudience` has a value ?

(I see that this was already the case in the previous version as well ?  https://docs.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.tokens.tokenvalidationparameters?view=azure-dotnet#properties)

src/Arcus.WebApi.Security/Model/ClaimCheck.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Collections.Generic;
+
+namespace Arcus.WebApi.Security.Model
+{
+    public class ClaimCheck : Dictionary<string, string>
+    {
+    }
```
fgheysels, line 7: In my opinion, `ClaimCheck` should be a simple value type.

Something that looks like 

```csharp
public class ClaimCheck
{
       public string Claim {get;}
       public string Value {get;}
}
```

Maybe even implement IEquatable on it.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
fgheysels: He's linking the claimchecks with the claims that exist in the token.   If the number of claims found that have the key and value of what is found in the claimchecks corresponds with the number of claimchecks, all checks have passed.

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
```diff
@@ -77,5 +78,33 @@ public void Constructor_WithTokenValidationParametersAndDiscoveryEndpoint_Succee
...
+        [Fact]
+        public void Constructor_WithoutClaimCheck_ThrowsException()
+        {
+            // Arrange
+            var claimCheck = new ClaimCheck();
+
+            // Act & Assert
+            Assert.Throws<ArgumentException>(() => new JwtTokenReader(claimCheck));
+        }
     }
```
fgheysels, line 109: It would be nice to have some tests which check if the functionality regarding the claim checking works.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
fgheysels: Actually, we could also choose to not set the  `ValidAudience` property, and rely on the claimcheck-check only ...

fgheysels (changes requested):

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: You are right @fgheysels. So basically I'm comparing the claims coming in ClaimCheck object vs the claims in the jwt object based on the claim name and value. If the result is the same number of claims the original ClaimCheck object, then all the claims are valid.

src/Arcus.WebApi.Security/Model/ClaimCheck.cs
egeda9: I also thought about this solution. However, a Dictionary gives us the implicit validation of having unique values. That is my only reason, otherwise, if it is a value type or not, the comparison logic will be the same, only a matter of definition of the object type.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: I thought about this option, however, I followed the proposal made in the bug, where the creation of a ClaimCheck object was mentioned. From my point of view, it is not necessary, I just followed that indication.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: Done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: Done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: Done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: Done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: Done

src/Arcus.WebApi.Security/Model/ClaimCheck.cs
fgheysels: You do have a point regarding the uniqueness, but I think that creating a type which inherits from Dictionary is just to implicit.   If you want to use a dictionary for this, then I'd suggest to pass in a Dictionary and not a type that inherits from Dictionary since that is not clear to the consumer of the library.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: done

src/Arcus.WebApi.Security/Model/ClaimCheck.cs
egeda9: done

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
egeda9: done

docs/features/security/auth/jwt.md
```diff
@@ -62,4 +62,55 @@ public class Startup
         });
     }
 }
+```
+
+### Custom Claim validation
+
+It allows validating not only not claims that are added to the authorization in the JWT token, but any type of custom claim that needs to be verified
```
fgheysels, line 69:
```suggestion
It allows validating not only on the audience claim in the JWT token, but any type of custom claim that needs to be verified
```

Double negation. I guess you meant this :)

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -25,12 +30,13 @@ public class JwtTokenReader : IJwtTokenReader
...
         /// <remarks>Uses Microsoft OpenId connect discovery endpoint</remarks>
-        /// <param name="applicationId">Azure AD Application used as audience to validate against</param>
-        public JwtTokenReader(string applicationId)
+        /// <param name="claimCheck">Azure AD Application used as audience to validate against</param>
+        public JwtTokenReader(IDictionary<string, string> claimCheck)
         {
-            Guard.NotNullOrWhitespace(applicationId, nameof(applicationId));
+            Guard.NotAny(claimCheck, nameof(claimCheck));
+
+            _claimCheck = claimCheck;
```
fgheysels, line 39: Don't we keep the existing constructor for backwards compatibility ?
The existing constructor which takes the `applicationId` can be implemented as such that it calls the JwtTokenReader claim that accepts the dictionary.

Something like this:

```csharp
public JwtTokenReader( string applicationId ) : this(new Dictionary { "aud" = applicationId })
{}
```
(Syntax errors are possible, but I think you get the idea :) )

src/Arcus.WebApi.Security/Authorization/Extensions/FilterCollectionExtensions.cs
```diff
@@ -23,9 +24,10 @@ public static FilterCollection AddJwtTokenAuthorization(this FilterCollection fi
         /// </summary>
         /// <param name="filters">All filters that are being applied to the request pipeline</param>
         /// <param name="configureOptions">Configuration options for using JWT token authorization</param>
+        /// <param name="claimCheck">Custom claims key-value pair to validate against</param>
         public static FilterCollection AddJwtTokenAuthorization(
             this FilterCollection filters, 
-            Action<JwtTokenAuthorizationOptions> configureOptions)
+            Action<JwtTokenAuthorizationOptions> configureOptions, IDictionary<string, string> claimCheck = null)
```
fgheysels, line 30: Can the `claimCheck` argument be null ?  What happens if we pass null ?

Hmm, seems like we're not doing anything with it ?

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -16,7 +19,9 @@ public class JwtTokenReader : IJwtTokenReader
         private const string MicrosoftDiscoveryEndpoint =
             "https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration";
 
-        private readonly string _applicationId;
+        private const string AudienceKeyName = "aud";
```
fgheysels, line 22: What about using the `JwtClaimTypes.Audience` constant which exists in [here](https://github.com/IdentityModel/IdentityModel/blob/main/src/JwtClaimTypes.cs)

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -63,29 +72,54 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
+            return _claimCheck.Count == validClaims;
         }
 
         private async Task<TokenValidationParameters> DetermineTokenValidationParametersAsync()
         {
             if (_tokenValidationParameters is null)
             {
-                OpenIdConnectConfiguration config = await _configManager.GetConfigurationAsync();
-                
+                var config = await _configManager.GetConfigurationAsync();
```
fgheysels, line 117: I think we tend to specify the type explicitly if we assign a method return value to it.
I'm sure @stijnmoreels can tell me if I'm right or wrong on this one

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -63,29 +72,54 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
+
                 var validationParameters = new TokenValidationParameters
                 {
-                    ValidateAudience = false,
-                    ValidAudience = _applicationId,
+                    ValidateAudience = _claimCheck.ContainsKey(AudienceKeyName),
+                    ValidAudience = _claimCheck[AudienceKeyName],
```
fgheysels, line 122: This will throw an exception if AudienceKeyName is not present in the _claimCheck dictionary.

(Would be good to create a testcase for this)

fgheysels (changes requested):

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
stijnmoreels: I was once told that we should limit the 'special' features in Arcus (was about `params`) so while I'm also a fan of `params`, maybe we should pass in an `IEnumerable<>` instead.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
stijnmoreels: Yes, absolutly necessary!

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
stijnmoreels: Yes, that right! The Arcus code guidelines specify that we should specify the type instead of using `var` if we don't know that the type is. So, in this case, it should be `OpenIdConnectionConfiguration`.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
stijnmoreels: Yes, a `Guard` would be sufficient here.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -63,29 +72,54 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
+        /// <param name="validatedToken">Security token which contains the jwt claims</param>
+        /// <returns></returns>
+        public bool ValidateClaimCheck(SecurityToken validatedToken)
+        {
+            if (_claimCheck == null || !_claimCheck.Any())
+            {
+                return true;
+            }
+
+            var jwtToken = (JwtSecurityToken) validatedToken;
```
stijnmoreels, line 103: A type cast exception can happen here. If we only allow such a type, we should `Guard` here.

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
```diff
@@ -77,5 +90,102 @@ public void Constructor_WithTokenValidationParametersAndDiscoveryEndpoint_Succee
...
+            // Assert
+            Assert.False(isValid);
+        }
+
+        private static string CreateJwt(string issuer, string authority, string symSec, int daysValid)
+        {
+            var tokenHandler = new JwtSecurityTokenHandler();
+            var claims = CreateClaimsIdentitiesAsync();
+
+            // Create JWToken
```
stijnmoreels, line 167:
```suggestion
```

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
```diff
@@ -77,5 +90,102 @@ public void Constructor_WithTokenValidationParametersAndDiscoveryEndpoint_Succee
...
+        {
+            // Arrange
+            var token = CreateJwt(Issuer, Authority, PrivateKey, DaysValid);
+            var handler = new JwtSecurityTokenHandler();
+            var securityToken = handler.ReadJwtToken(token);
+
+            var claimCheck = new Dictionary<string, string>
+            {
+                {"aud", Issuer},
+                {"oid", "fa323e12-e4b8-4e22-bb2a-b18cb4b76301"},
```
stijnmoreels, line 149: These checks have the same ID's as the ones that are created for the identities. It should be good that these are both be generated instead of hard-coded like this.

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
```diff
@@ -7,24 +12,32 @@ namespace Arcus.WebApi.Tests.Unit.Security.Authorization
 {
     public class JwtTokenReaderTests
     {
+        private const string Issuer = "https://myut.com";
+        private const string Authority = "https://myut.com";
+        private const string PrivateKey = "J6k2eVCTXDp5b97u6gNH5GaaqHDxCmzz2wv3PRPFRsuW2UavK8LGPRauC4VSeaetKTMtVmVzAC8fh8Psvp8PFybEvpYnULHfRpM8TA2an7GFehrLLvawVJdSRqh2unCnWehhh2SJMMg5bktRRapA8EGSgQUV8TCafqdSEHNWnGXTjjsMEjUpaxcADDNZLSYPMyPSfp6qe5LMcd5S9bXH97KeeMGyZTS2U8gp3LGk2kH4J4F3fsytfpe9H9qKwgjb";
```
stijnmoreels, line 17: I hope this is not sensitive information and just test value, but if it is test value, I'm wondering why it's hard-coded and not generated.

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
```diff
@@ -77,5 +90,102 @@ public void Constructor_WithTokenValidationParametersAndDiscoveryEndpoint_Succee
...
+            // Act
+            var jwtTokenReader = new JwtTokenReader(claimCheck);
+            var isValid = jwtTokenReader.ValidateClaimCheck(securityToken);
+
+            // Assert
+            Assert.True(isValid);
+        }
+
+        [Fact]
+        public async Task Constructor_ValidateClaimCheck_Invalid()
```
stijnmoreels, line 139: This test talks about invalid claim check, but is a _almost_ the same as the success version. It should be good that the difference is clear as to _why_ the claim check fails. (_becasue the claim check identities doesn't match_).

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
```diff
@@ -77,5 +90,102 @@ public void Constructor_WithTokenValidationParametersAndDiscoveryEndpoint_Succee
...
+
+            // Act
+            var jwtTokenReader = new JwtTokenReader(tokenValidationParameters, openIdConnectDiscoveryUri, claimCheck);
+
+            // Assert
+            Assert.NotNull(jwtTokenReader);
+        }
+
+        [Fact]
+        public async Task Constructor_ValidateClaimCheck_Succeeds()
```
stijnmoreels, line 116: I would like to see some tests with more boundary test cases. Empty claim checks in both the configureed as the incoming, same identities but different claim check values, claim check with empty claim check values (should throw here), ... Now it's fairly limited as to what's being tested.

src/Arcus.WebApi.Security/Authorization/Extensions/FilterCollectionExtensions.cs
stijnmoreels: Yes, some tests on this extension and not on the reader would be good too. See if the extension not only guards against invalid input but also configured the reader correctly.

stijnmoreels (changes requested):
Were getting there, thanks for the work!

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
tomkerkhove: It's not that we limit it because it's "special" but more how we consume it, I prefer a list over arrays as they are easier to use but if you both think params is better then that's fine for me.

stijnmoreels: Yeah, sorry, my bad, 😅 , in this case, in a constructor, I don't think `params` is something that's helpful. It's more something you see in methods I think. (although I'm guilty of using it in ctors).

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
fgheysels: A Guard will not solve the issue.  
The fact is , that checking the audience is optional, and should only happen when there's a claimcheck defined for this.

Code should be something like this:

```csharp
ValidateAudience = _claimChecks.ContainsKey(AudienceKeyName);
ValidAudience = (_claimChecks.ContainsKey(AudienceKeyName) ? _claimChecks[AudienceKeyName] : null
```

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
egeda9: done

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
egeda9: done

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
egeda9: done

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenReaderTests.cs
egeda9: done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: done

docs/features/security/auth/jwt.md
egeda9: done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -25,12 +29,13 @@ public class JwtTokenReader : IJwtTokenReader
         ///     Constructor
         /// </summary>
         /// <remarks>Uses Microsoft OpenId connect discovery endpoint</remarks>
-        /// <param name="applicationId">Azure AD Application used as audience to validate against</param>
-        public JwtTokenReader(string applicationId)
+        /// <param name="claimCheck">Azure AD Application used as audience to validate against</param>
+        public JwtTokenReader(IDictionary<string, string> claimCheck)
         {
-            Guard.NotNullOrWhitespace(applicationId, nameof(applicationId));
+            Guard.NotAny(claimCheck, nameof(claimCheck));
```
fgheysels, line 35: Do you need to guard here on claimChecks for which the value is empty ?

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -39,21 +44,33 @@ public JwtTokenReader(string applicationId)
...
+        {
+        }
+
+        /// <summary>
+        ///     Constructor
+        /// </summary>
+        /// <param name="applicationId">Azure AD Application used as audience to validate against</param>
+        public JwtTokenReader(string applicationId) : this(new Dictionary<string, string> {{ AudienceKeyName, applicationId}})
         {
+            Guard.NotNullOrWhitespace(applicationId, nameof(applicationId));
```
fgheysels, line 58: If you Guard on empty claimchecks (claimchecks where the value is not specified) in the other constructor, then this Guard here can be removed I guess

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -16,7 +18,9 @@ public class JwtTokenReader : IJwtTokenReader
         private const string MicrosoftDiscoveryEndpoint =
             "https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration";
 
-        private readonly string _applicationId;
+        private const string AudienceKeyName = JwtClaimTypes.Audience;
```
fgheysels, line 21: Minor remark, but this private member can be removed if you use JwtClaimTypes.Audience directly instead of the `AudienceKeyName` member

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -63,29 +80,54 @@ public JwtTokenReader(TokenValidationParameters tokenValidationParameters, strin
...
+            }
+
+            JwtSecurityToken jwtToken = (JwtSecurityToken) validatedToken;
+
+            int validClaims = (from c in jwtToken.Claims
+                join m in _claimCheck on c.Type equals m.Key
+                where c.Value == m.Value
+                select c).Distinct().Count();
+
+            return _claimCheck.Count == validClaims;
```
stijnmoreels, line 118: I was wondering, just curious, why this approach wasn't used:

```csharp
return jwtToken.Claims.All(c => _claimCheck.TryGetValue(c.Type, out string value) && value == c.Value);
```

stijnmoreels:
OK, I think we're now closing in on the basic implementation. Maybe now we can add some integration tests (sorry, their still mostly in the unit test project), for this claim check.
I'm thinking about tests where the app describes more or less claim checks, empty, or where the incoming JWT has defined.

tomkerkhove:
62 comments on this PR, is everything clear @egeda9 or do you prefer if we setup a small call to unblock you?

Thanks for your contribution btw!

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: The method you mention is a super method of:

```csharp 

public JwtTokenReader (string applicationId): this (new Dictionary <string, string> {{JwtClaimTypes.Audience, applicationId}})

```
therefore the validation is pertinent as long as the method is called with the corresponding signature (parameters). In case it is called through the reference of the base method, it is verified that the applicationid is not null or empty in the respective method

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: Beyond validation, what I would like to verify with you at this point is whether a claim can be empty or not. If it is verified in the list of claims that none should be empty, it is assumed that the logic does not allow any claim with an empty or null value. What I do not know is whether in a real scenario this situation can occur

src/Arcus.WebApi.Security/Authorization/Extensions/FilterCollectionExtensions.cs
egeda9: done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: done

egeda9:
Guys, I just pushed some new Unit Tests regarding the suggestions that I got from the comments above. For now, I think every suggestion was answered or implemented in code. So please take a look and let me know your thoughts with these new changes.

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenAutorizationFilterTests.cs
```diff
@@ -56,6 +55,34 @@ public async Task GetHealthWithCorrectBearerToken_WithAzureManagedIdentityAuthor
             }
         }
 
+        [Fact]
+        public async Task GetHealthWithCorrectBearerToken_WithAzureManagedIdentityAuthorizationAndCustomClaims_ReturnsOk()
```
stijnmoreels, line 59: We could benefit from some failure tests too. See how the header name, the amount of claims... affect the result.

stijnmoreels (changes requested):

src/Arcus.WebApi.Security/Arcus.WebApi.Security.csproj
```diff
@@ -28,6 +28,7 @@
   <ItemGroup>
     <PackageReference Include="Arcus.Security.Providers.AzureKeyVault" Version="1.3.0" />
     <PackageReference Include="Guard.Net" Version="1.2.0" />
+    <PackageReference Include="IdentityModel" Version="3.10.10" />
```
tomkerkhove, line 31: This is a very old version, can you update it please? Or don't we need it anymore?

docs/features/security/auth/jwt.md
```diff
@@ -62,4 +62,55 @@ public class Startup
...
+```
+
+### Custom Claim validation
+
+It allows validating not only on the audience claim in the JWT token, but any type of custom claim that needs to be verified
+
+```csharp
+public class Startup
+{
+    public void ConfigureServices(IServiceCollection services)
```
tomkerkhove, line 74: Based on https://github.com/arcus-azure/arcus.webapi/pull/179/files#diff-48f283b1ac9786c91d89d9bb04fcc05cf8d48411eb1086840a67a7de9dcf98cfR33 this looks outdated, is that correct?

tomkerkhove (approved):
Just some minor nits regarding docs, but LGTM for the rest!

src/Arcus.WebApi.Tests.Unit/Security/Authorization/JwtTokenAutorizationFilterTests.cs
tomkerkhove: Let's do a follow-up PR for more tests!

tomkerkhove:
Oh and can you have a look at the failing tests @egeda9? Thanks a ton!

src/Arcus.WebApi.Security/Authorization/Extensions/FilterCollectionExtensions.cs
```diff
@@ -23,13 +27,14 @@ public static FilterCollection AddJwtTokenAuthorization(this FilterCollection fi
...
```
stijnmoreels, line 33: We may consider creating an overload instead of an optional parameter because it will result in `MethodNotFound` exceptions when the packages are updated seperatly.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -39,21 +43,33 @@ public JwtTokenReader(string applicationId)
         /// </summary>
         /// <remarks>Uses Microsoft OpenId connect discovery endpoint</remarks>
         /// <param name="tokenValidationParameters">Collection of parameters to influence how the token validation is done</param>
-        public JwtTokenReader(TokenValidationParameters tokenValidationParameters) : this(tokenValidationParameters, MicrosoftDiscoveryEndpoint)
+        /// <param name="claimCheck">Custom claims key-value pair to validate against</param>
+        public JwtTokenReader(TokenValidationParameters tokenValidationParameters, IDictionary<string, string> claimCheck = null) : this(tokenValidationParameters, MicrosoftDiscoveryEndpoint, claimCheck)
```
stijnmoreels, line 47: We may consider a constructor overload instead of an optional parameter because it will result in `MethodNotFound` exceptions when the packages are updated sepeartly.

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
```diff
@@ -39,21 +43,33 @@ public JwtTokenReader(string applicationId)
...
+        }
+
         /// <summary>
         ///     Constructor
         /// </summary>
         /// <param name="openIdConnectDiscoveryUri">Uri of an OpenId connect endpoint for discovering the configuration</param>
         /// <param name="tokenValidationParameters">Collection of parameters to influence how the token validation is done</param>
-        public JwtTokenReader(TokenValidationParameters tokenValidationParameters, string openIdConnectDiscoveryUri)
+        /// <param name="claimCheck">Custom claims key-value pair to validate against</param>
+        public JwtTokenReader(TokenValidationParameters tokenValidationParameters, string openIdConnectDiscoveryUri, IDictionary<string, string> claimCheck = null)
```
stijnmoreels, line 66: We may consider creating an method overload instead of an optional parameter because it will result in `MethodNotFound` exceptions when the packages are updated seperatly.

src/Arcus.WebApi.Security/Authorization/JwtTokenAuthorizationOptions.cs
```diff
@@ -36,6 +37,15 @@ public JwtTokenAuthorizationOptions()
         {
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="JwtTokenAuthorizationOptions"/> class.
+        /// </summary>
+        /// <param name="claimCheck">Custom claims key-value pair to validate against.</param>
+        public JwtTokenAuthorizationOptions(IDictionary<string, string> claimCheck = null)
```
stijnmoreels, line 44: We may consider an constructor overload instead of an optional parameter because it will result in `MethodNotFound` exceptions when the packages are updated seperatly.

stijnmoreels (changes requested):

stijnmoreels:
When the tests are succeeding and the optional parameter is reconsidered, we can complete this, I guess. Thanks for sticking arround! 👍

src/Arcus.WebApi.Security/Arcus.WebApi.Security.csproj
egeda9: done

docs/features/security/auth/jwt.md
egeda9: this is the latest version, nothing has changed here

src/Arcus.WebApi.Security/Authorization/Extensions/FilterCollectionExtensions.cs
egeda9: done

src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
egeda9: done

src/Arcus.WebApi.Security/Authorization/JwtTokenAuthorizationOptions.cs
egeda9: done

egeda9:
Hi guys, regarding the unit tests I found the issue in the pipeline. https://dev.azure.com/codit/Arcus/_build/results?buildId=49293&view=logs&j=6b58850f-3858-5a05-33e2-5e41cbf03c4e&t=67dc9284-07ce-51a8-a760-4b4b8180fb92&l=413

`System.PlatformNotSupportedException : Operation is not supported on this platform.
  Stack Trace:
     at System.Security.Cryptography.RSA.ToXmlString`

I think you guys need to update the target framework for the unit test project in the deployment file to net core 3.1, since I'm using it to generate the private key to create the dummy JWT token. 

Here I found more details: https://github.com/dotnet/core/issues/874

src/Arcus.WebApi.Tests.Unit/Hosting/TestOpenIdServer.cs
```diff
@@ -83,10 +103,47 @@ public async Task<string> RequestAccessTokenAsync()
...
+        /// <summary>
+        /// Requests a JWT bearer access token to send authorized HTTP requests.
+        /// </summary>
+        public string RequestSecretToken(string issuer, string authority, string symSec, int daysValid, IDictionary<string, string> claims = null)
+        {
+            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
+            ClaimsIdentity claimsIdentity = null;
+
+            if (claims != null)
+                claimsIdentity = CreateClaimsIdentitiesAsync(claims);
```
stijnmoreels, line 119:
```suggestion
            if (claims != null)
            {
                claimsIdentity = CreateClaimsIdentitiesAsync(claims);
            }    
```

src/Arcus.WebApi.Tests.Unit/Hosting/TestOpenIdServer.cs
```diff
@@ -83,10 +103,47 @@ public async Task<string> RequestAccessTokenAsync()
...
+                Issuer = issuer,
+                SigningCredentials = new SigningCredentials(new SymmetricSecurityKey(Encoding.Default.GetBytes(symSec)), SecurityAlgorithms.HmacSha256Signature)
+            };
+
+            JwtSecurityToken token = tokenHandler.CreateJwtSecurityToken(tokenDescriptor);
+
+            return tokenHandler.WriteToken(token);
+        }
+
+        private static ClaimsIdentity CreateClaimsIdentitiesAsync(IDictionary<string, string> claims)
```
stijnmoreels, line 135: Why is this called `...Async`? 😄

stijnmoreels (changes requested):

stijnmoreels:
> Hi guys, regarding the unit tests I found the issue in the pipeline. https://dev.azure.com/codit/Arcus/_build/results?buildId=49293&view=logs&j=6b58850f-3858-5a05-33e2-5e41cbf03c4e&t=67dc9284-07ce-51a8-a760-4b4b8180fb92&l=413
> 
> `System.PlatformNotSupportedException : Operation is not supported on this platform. Stack Trace: at System.Security.Cryptography.RSA.ToXmlString`
> 
> I think you guys need to update the target framework for the unit test project in the deployment file to net core 3.1, since I'm using it to generate the private key to create the dummy JWT token.
> 
> Here I found more details: [dotnet/core#874](https://github.com/dotnet/core/issues/874)

Hmm, we support both 2.2 as 3.1 target frameworks in the unit tests and load both target frameworks during the test run. Is their any way arround this bc it seems a bit annoying that we would change the target framework (or only run this test in the correct target framework), for a generation line.

src/Arcus.WebApi.Tests.Unit/Hosting/TestOpenIdServer.cs
egeda9: fixed

src/Arcus.WebApi.Tests.Unit/Hosting/TestOpenIdServer.cs
egeda9: fixed

egeda9:
Hi guys, all comments resolved, and also unit tests running successfully. Please let me know if you have any questions or suggestions.

tomkerkhove (approved):
LGTM, thanks a ton for the contribution!

src/Arcus.WebApi.Tests.Unit/Hosting/TestOpenIdServer.cs
```diff
@@ -138,7 +197,8 @@ private static void ConfigureServices(IServiceCollection services)
                             AllowedScopes = { "api1" }
                         }
                     })
-                    .AddDeveloperSigningCredential();
+                    //.AddDeveloperSigningCredential()
```
stijnmoreels, line 200: Can be removed if we don't use it anymore.

stijnmoreels (approved):

stijnmoreels:
Thanks for the great work!

fgheysels (approved):

src/Arcus.WebApi.Tests.Unit/Hosting/TestOpenIdServer.cs
tomkerkhove:
```suggestion
```

tomkerkhove:
Can we just merge master @egeda9 ?

egeda9:
Hi guys, not sure what is happening, but I am unable to merge this branch to master

image

Regarding the conflicting files, again, not sure why but everything is up to date

image

Could you guys please verify if there is some kind of permissions missing here?

tomkerkhove:
YOu need to pull in master branch from this repo, not your master branch of your fork.
But if you want, @stijnmoreels did some decent changes to the setup so we could merge your PR in a temp branch and have Stijn migrate it since he knows how it changed.

Do you preference that?

egeda9:
@tomkerkhove  yes perfect, please merge this PR by using @stijnmoreels help. Thank you

tomkerkhove:
Thanks @egeda9! It's all yours @stijnmoreels, do your ✨

PR #176, opened 2020-09-07
Feature - revisit API security filters on input formatting, execution order, and logging

stijnmoreels:
Revisiting the API security filters in the web API repo.
- Adding extra BASE64 format check in certificate authentication filter in favor of throwing an `FormatException` in `Convert.FromBase64String`
- Adding extra BASE64URL checks on the JWT in the JWT authentication filter before passing along to custom JWT reader (format validation before signature verification, and the reader is considered here to be responsible for the signature check)
- Moving required services in the filters as first task to do to make sure that the reason for throwing exceptions is first checked becore checking if headers/query/... values exists (so configuration failures before authentication failures)
- Adding logging throughout the security filters so we can follow the path

PR #185, opened 2020-10-19
Feature - bypass API security filters with attributes

### LINKED ISSUE #184: Bypass JwtTokenValidation for certain endpoints
In a project I'm working on, we're making use of the `JwtTokenAuthorization` filter to verify if all incoming requests are coming via Azure API Management.   This works fine, but I was wondering if there is a way to bypass this validation / filter for certain endpoints.

Our API has a health endpoint, and this endpoint is called at regular intervals by Azure for checking if our API is still online.   However, due to the JwtTokenAuthorization filter that is configured, calls to this endpoint by Azure always return a HTTP 401.

Therefore, I was wondering if there are possibilities to bypass this filter for certain endpoints.

stijnmoreels:
Provide the capability to bypass authentication and authorization with specific attributes and with the default `AllowAnonymous` attribute.

> Maybe it's good to merge later with #176 because it also adds logging to the API filters + provides an extension to safely retrieve a logger.

Closes #184

### DIFF
--- a/docs/preview/features/security/auth/certificate.md
+++ b/docs/preview/features/security/auth/certificate.md
@@ -130,7 +130,7 @@ public class SystemController : ControllerBase
 {
     [HttpGet]
     [BypassCertificateAuthentication]
-    public IActionResult Get()
+    public IActionResult GetHealth()
     {
         return Ok();
     }
--- a/docs/preview/features/security/auth/certificate.md
+++ b/docs/preview/features/security/auth/certificate.md
@@ -126,7 +126,7 @@ This works with adding one of these attributes to the respectively endpoint:
 ```csharp
 [ApiController]
 [CertificateAuthentication]
-public class MyController : ControllerBase
+public class SystemController : ControllerBase
 {
     [HttpGet]
     [BypassCertificateAuthentication]
--- a/docs/preview/features/security/auth/certificate.md
+++ b/docs/preview/features/security/auth/certificate.md
@@ -128,7 +128,7 @@ This works with adding one of these attributes to the respectively endpoint:
 [CertificateAuthentication]
 public class SystemController : ControllerBase
 {
-    [HttpGet]
+    [HttpGet('api/v1/health')]
     [BypassCertificateAuthentication]
     public IActionResult GetHealth()
     {
--- a/docs/preview/features/security/auth/certificate.md
+++ b/docs/preview/features/security/auth/certificate.md
@@ -16,6 +16,7 @@ This authentication process consists of following parts:
 The package allows two ways to configure this type of authentication mechanism in an <span>ASP.NET</span> application:
 - [Globally enforce certificate authentication](#Globally-enforce-certificate-authentication)
 - [Enforce certificate authentication per controller or operation](#Enforce-certificate-authentication-per-controller-or-operation)
+- [Bypassing authentication](#bypassing-authentication)
 
 ## Installation
 
@@ -112,5 +113,29 @@ public class MyApiController : ControllerBase
     }
 }
 ```
+
+## Bypassing authentication
+
+The package supports a way to bypass the certificate authentication for certain endponts.
+This works with adding one of these attributes to the respectively endpoint:
+- `BypassCertificateAuthentication`
+- `AllowAnonymous`
+
+> Works on both method and controller level, using either the certificate filter or attribute.
+
+```csharp
+[ApiController]
+[CertificateAuthentication]
+public class MyController : ControllerBase
+{
+    [HttpGet]
+    [BypassCertificateAuthentication]
+    public IActionResult Get()
+    {
+        return Ok();
+    }
+}
+```
+
 
 [&larr; back](/)
--- a/docs/preview/features/security/auth/jwt.md
+++ b/docs/preview/features/security/auth/jwt.md
@@ -11,7 +11,12 @@ This authorization process consists of the following parts:
 1. Find the OpenID server endpoint to request the correct access token
 2. Determine the request header name you want to use where the access token should be available
 
-## Installation
+- [Globally enforce JWT authorization](#globally-enforce-jwt-authorization)
+- [Bypassing authorization](#bypassing-authorization)
+
+## Globally enforce JWT authorization
+
+### Installation
 
 This feature requires to install our NuGet package:
 
@@ -19,7 +24,7 @@ This feature requires to install our NuGet package:
 PM > Install-Package Arcus.WebApi.Security
 ```
 
-## Usage
+### Usage
 
 The `JwtTokenAuthorizationFilter` can be added to the request filters in an <span>ASP.NET</span> Core application.
 This filter will then add authorization to all endpoints via the configured properties on the filter itself.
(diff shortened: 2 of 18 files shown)


fgheysels:
Looks good to me at first sight!  Thx.  Haven't done a thorough review yet however

docs/preview/features/security/auth/certificate.md
```diff
@@ -113,4 +114,28 @@ public class MyApiController : ControllerBase
...
+This works with adding one of these attributes to the respectively endpoint:
+- `BypassCertificateAuthentication`
+- `AllowAnonymous`
+
+> Works on both method and controller level, using either the certificate filter or attribute.
+
+```csharp
+[ApiController]
+[CertificateAuthentication]
+public class MyController : ControllerBase
```
tomkerkhove, line 129:
```suggestion
public class SystemController : ControllerBase
```

tomkerkhove: Let's use a more concrete example

docs/preview/features/security/auth/certificate.md
```diff
@@ -113,4 +114,28 @@ public class MyApiController : ControllerBase
...
+{
+    [HttpGet]
+    [BypassCertificateAuthentication]
+    public IActionResult Get()
```
tomkerkhove, line 133:
```suggestion
    public IActionResult GetHealth()
```

docs/preview/features/security/auth/certificate.md
```diff
@@ -113,4 +114,28 @@ public class MyApiController : ControllerBase
...
+- `AllowAnonymous`
+
+> Works on both method and controller level, using either the certificate filter or attribute.
+
+```csharp
+[ApiController]
+[CertificateAuthentication]
+public class MyController : ControllerBase
+{
+    [HttpGet]
```
tomkerkhove, line 131:
```suggestion
    [HttpGet('api/v1/health')]
```

docs/preview/features/security/auth/jwt.md
```diff
@@ -62,4 +67,26 @@ public class Startup
...
+
+The package supports a way to bypass the JWT authorization for certain endponts.
+This works with adding one of these attributes to the respectively endpoint:
+- `BypassJwtAuthorization`
+- `AllowAnonymous`
+
+> Works on both method and controller level.
+
+```csharp
+[ApiController]
```
tomkerkhove, line 82: See feedback other example

docs/preview/features/security/auth/shared-access-key.md
```diff
@@ -112,4 +114,26 @@ public void ConfigureServices(IServiceCollections services)
...
 
+## Bypassing authentication
+The package supports a way to bypass the shared access key authentication for certain endponts.
+This works with adding one of these attributes to the respectively endpoint:
+- `BypassSharedAccessKeyAuthentication`
+- `AllowAnonymous`
+
+> Works on both method and controller level, using either the shared access key filter or attribute.
+
+```csharp
```
tomkerkhove, line 125: See feedback other example

src/Arcus.WebApi.Security/Authentication/Certificates/BypassCertificateAuthenticationAttribute.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Arcus.WebApi.Security.Authentication.Certificates
+{
+    /// <summary>
+    /// Attribute to circumvent the certificate authentication.
+    /// </summary>
+    public class BypassCertificateAuthenticationAttribute : Attribute
```
tomkerkhove, line 8: Do we actually need an attribute per scenario or do we create a more generic one?

/cc @fgheysels

tomkerkhove (approved):
LGTM, just one point to clear out with @fgheysels

src/Arcus.WebApi.Security/Authentication/Certificates/BypassCertificateAuthenticationAttribute.cs
fgheysels: I think it is a good idea to have a specific attribute to bypass the jwt token authentication, as this is used for backdoor security.
I'm wondering if it just isn't enough to specify the `AllowAnonymous` certificate to bypass certificate based authentication ?  (I'm thinking this has more to do with front-door security ? )

stijnmoreels: Yeah, I get that. Maybe we should not include this allow anonymous attribute in our authentication filters.

tomkerkhove: I think we should have a unified approach for all options as it depends on the scenario, no?

I'm not sure if I get what you are saying 🤔

tomkerkhove: Why not? It's a commonly used attribute so I'm not sure why we should not support this? :think

stijnmoreels: Yeah but anonymous is maybe more an authorization term than an authentication term.

tomkerkhove: Here's how .NET Core positions the usage of `AllowAnonymous`:
image

https://docs.microsoft.com/en-us/aspnet/core/security/authorization/simple?view=aspnetcore-3.1

This feels similar to what we are adding, no?

stijnmoreels: Yes, and it says "bypasses all authorization statements", it's a thin line of course.

fgheysels: So, I would think 'AllowAnonymous' is ok instead of those specific attributes.
(Although I'd like to not use AllowAnonymous to bypass the JwtToken validation.   We might have an endpoint that requires authentication for instance, but doesn't  require backdoor security).
Or am I exaggerating ?

tomkerkhove: Then we add both so you can pick and choose, ok?

fgheysels: You mean that applying `AllowAnonymous` will also automatically bypass the backdoor security ?

tomkerkhove: Yes, but if you prefer to use the specific attribute you can still have the JWT token validation for that specific operation.

Those who want to allow anonymous access on, for example a health endpoint, can use AllowAnonymous

fgheysels: Ok, I can agree.
Then do we also keep the BypassCertificateAuthenticationAttribute, or is this one completely being replaced by the common AllowAnonymous ?

tomkerkhove: I'd keep those for your scenarios where you only want to bypass certificate auth and not JWT.

But if you say it's no longer required then we should leave it out I think

tomkerkhove: Discussed this with @fgheysels and we've agreed to support `AllowAnonymous` and have a `Bypass{Auth Type}` attribute/filter for more granular control

fgheysels (approved):
I think this is OK. Thx.

PR #186, opened 2020-10-26
Feature - add behavior to allow custom claims validation (#175)

### LINKED ISSUE #175: Allow optional jwt token validation using the oid

stijnmoreels:
@egeda9's PR now merged with `master`.

Closes #175 
Relates to #179


stijnmoreels:
Thanks @egeda9 for your hard work! 🎉

docs/features/security/auth/jwt.md
```diff
@@ -62,4 +62,55 @@ public class Startup
         });
     }
 }
+```
+
+### Custom Claim validation
+
+It allows validating not only on the audience claim in the JWT token, but any type of custom claim that needs to be verified
+
+```csharp
```
tomkerkhove, line 71: This looks outdated based on the code, no? Ideally you don't have to pass the claims to the reader which is the case I think?

tomkerkhove (changes requested):
Just needs a doc update and we are good to go!

docs/features/security/auth/jwt.md
stijnmoreels: And additionally, this should be in the preview docs, right 🤔 . Not sure how this slipped the first review 😅 .

stijnmoreels: Ah, and this additional section is purely about the claims, the section above talks about options where you don't have to pass the claim check. So, it's good, I guess?

tomkerkhove: I don't think so because based on the extension code you don't pass the claims to the reader but the extension itself?

stijnmoreels: Aha! Now, I see it.

stijnmoreels: I'll also update the extensions so you're not required to pass along the configure options function every time and only can pass the claim checks (+ update the docs of course).

tomkerkhove (approved):

PR #196, opened 2021-02-12
Feature - use `.LogSecurityEvent` for security types

stijnmoreels:
Now that we have extracted the SQL package in the Observability repo (https://github.com/arcus-azure/arcus.observability/issues/131), we can use the `.LogSecurityEvent` instead.

### DIFF
--- a/src/Arcus.WebApi.Security/Arcus.WebApi.Security.csproj
+++ b/src/Arcus.WebApi.Security/Arcus.WebApi.Security.csproj
@@ -26,7 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Arcus.Security.Providers.AzureKeyVault" Version="1.3.0" />
+    <PackageReference Include="Arcus.Security.Providers.AzureKeyVault" Version="1.4.0" />
     <PackageReference Include="Guard.Net" Version="1.2.0" />
     <PackageReference Include="IdentityModel" Version="4.4.0" />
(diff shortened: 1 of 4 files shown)


src/Arcus.WebApi.Security/Arcus.WebApi.Security.csproj
```diff
@@ -22,15 +22,13 @@
...
   <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp3.1'">
     <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.9" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Arcus.Security.Providers.AzureKeyVault" Version="1.3.0" />
+    <PackageReference Include="Arcus.Security.Providers.AzureKeyVault" Version="1.4.0" />
     <PackageReference Include="Guard.Net" Version="1.2.0" />
     <PackageReference Include="IdentityModel" Version="4.4.0" />
```
stijnmoreels, line 30: We only use the enumuration `ClaimTypes` in this package. Is that enough to be a dependency here? Can't we use our own enum then?

tomkerkhove: Why have our own if there is somebody who maintains it? `IdentityModel` is maintained by IdentityServer which has a big userbase and knows more about security than us, so pretty sure they'll do a better job than us + it's compatible with users who build on top of their frameworks as well.

tomkerkhove (approved):

PR #197, opened 2021-02-23
fix: Ensure request body is tracked

### LINKED ISSUE #165: UseRequestTracking  with IncludeBodyRequest throws an exception

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

Ensure the request body is tracked if need be, and do it before running any other middleware so that the stream cannot be disposed of already.

That's the root cause of our problem where it tries to get the stream position, but somebody else already disposed of it (which is out of Arcus control, so we have to coop with it).

Closes #165

### DIFF
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -62,11 +62,11 @@ namespace Arcus.WebApi.Logging
             finally
             {
                 stopwatch.Stop();
-                await TrackRequestAsync(requestBody, httpContext, stopwatch.Elapsed);
+                TrackRequest(requestBody, httpContext, stopwatch.Elapsed);
             }
         }
 
-        private async Task TrackRequestAsync(string requestBody, HttpContext httpContext, TimeSpan duration)
+        private void TrackRequest(string requestBody, HttpContext httpContext, TimeSpan duration)
         {
             try
             {
@@ -145,7 +145,7 @@ namespace Arcus.WebApi.Logging
                 return "Request body could not be tracked because stream is not seekable";
             }
 
-            string contents = null;
+            string contents;
             long? originalPosition = null;
 
             try
@@ -159,9 +159,11 @@ namespace Arcus.WebApi.Logging
                 var reader = new StreamReader(requestStream);
                 contents = await reader.ReadToEndAsync();
             }
-            catch (Exception ex)
+            catch
             {
-                _logger.LogWarning(ex, "Unable to get request body for request tracking");
+                // We don't want to track additional telemetry for cost purposes,
+                // so we surface it like this
+                contents = "Unable to get request body for request tracking";
             }
 
             try
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -47,9 +47,12 @@ namespace Arcus.WebApi.Logging
         public async Task Invoke(HttpContext httpContext)
         {
             var stopwatch = Stopwatch.StartNew();
+
+            string requestBody = null;
             if (_options.IncludeRequestBody)
             {
                 httpContext.Request.EnableBuffering();
+                requestBody = await GetRequestBodyAsync(httpContext);
             }
 
             try
@@ -59,18 +62,17 @@ namespace Arcus.WebApi.Logging
             finally
             {
                 stopwatch.Stop();
-                await TrackRequestAsync(httpContext, stopwatch.Elapsed);
+                await TrackRequestAsync(requestBody, httpContext, stopwatch.Elapsed);
             }
         }
 
-        private async Task TrackRequestAsync(HttpContext httpContext, TimeSpan duration)
+        private async Task TrackRequestAsync(string requestBody, HttpContext httpContext, TimeSpan duration)
         {
             try
             {
                 IDictionary<string, StringValues> telemetryContext = GetRequestHeaders(httpContext);
 
-                string requestBody = await GetRequestBodyAsync(httpContext);
-                if (requestBody != null)
+                if (string.IsNullOrWhiteSpace(requestBody) == false)
                 {
                     telemetryContext.Add("Body", requestBody);
                 }
@@ -131,7 +133,7 @@ namespace Arcus.WebApi.Logging
             return requestHeaders.Where(header => _options.OmittedHeaderNames.Contains(header.Key) == false);
         }
 
-        private static async Task<string> SanitizeRequestBodyAsync(Stream requestStream)
+        private async Task<string> SanitizeRequestBodyAsync(Stream requestStream)
         {
             if (!requestStream.CanRead)
             {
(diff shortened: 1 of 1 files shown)


src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -143,18 +145,31 @@ private static async Task<string> SanitizeRequestBodyAsync(Stream requestStream)
...
+                }
 
-            var reader = new StreamReader(requestStream);
-            string contents = await reader.ReadToEndAsync();
+                var reader = new StreamReader(requestStream);
+                contents = await reader.ReadToEndAsync();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning(ex, "Unable to get request body for request tracking");
```
tomkerkhove, line 164: I figured to write a warning but happy to remove if need be @fgheysels

tomkerkhove:
I was also encountering this exception, just like you  @fgheysels:

> [10:51:16 FTL] Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead.
System.InvalidOperationException: Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.Read(Byte[] buffer, Int32 offset, Int32 count)

Can be fixed by following:
```
services.Configure<KestrelServerOptions>(options =>
            {
                options.AllowSynchronousIO = true;
            });
```

https://stackoverflow.com/questions/47735133/asp-net-core-synchronous-operations-are-disallowed-call-writeasync-or-set-all

But that felt like a hack. After this fix, that issue was gone as well.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
fgheysels: I don't know what I 'd do.  Log it as a wzrning, or log that in the 'requestbody' custom property ....

fgheysels:
Thanks for investigating!
The good news is that it now works:  the request-body is logged.
The bad news is however, that the 'request timeline' is a bit messed up now.

This is how my request timeline looks like if I do not use request-tracking:

image

When I do use request tracking, the timeline looks like this:

image

As you can see, now the request to my API endpoint is represented twice, and it's only the last request which contains the request-body.  (So , the one that is added extra as opposed to the first screenshot).

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
fgheysels: (The latter one would be in line with the other loggings)

tomkerkhove:
Doesn't look ideal, but I imagine you'd have the same if you turn of the request body tracking, or not? Otherwise I'd say we open a new issue.

Where does this run? Locally or on Azure? If it's on Azure, does it have the Application Insights extension installed?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: So in that case, you would add the text field in the custom property instead then?

fgheysels:
> Doesn't look ideal, but I imagine you'd have the same if you turn of the request body tracking, or not? Otherwise I'd say we open a new issue.
> 
> Where does this run? Locally or on Azure? If it's on Azure, does it have the Application Insights extension installed?

When I don't have the request body tracking enabled, I don't have that issue (first screenshot).
It runs on Azure.  Which AppInsights extensions are you referring to ?

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels, line 164:
```suggestion
            catch (Exception exception)
            {
                _logger.LogWarning(exception, "Unable to get request body for request tracking");
```

Something we do in Arcus, right? Full names except for lamdba's. 😉

stijnmoreels (approved):
👍

stijnmoreels:
Is this something we can explicitly test @tomkerkhove , to make sure that we did it right? As I remember correctly, it was first an issue of reproduction this, so could be a valuable test.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: I don't like abbreviations :D

tomkerkhove:
It is still an issue :D I've moved Arcus to our solution to reproduce it, didn't manage to do it in Arcus.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
tomkerkhove: Done

Issue #201, opened 2021-02-26
Extract the request tracking functionality into the `Arcus.WebApi.Logging.Core` so it can be re-used within Azure Functions


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Just like we extracted the HTTP correlation, it could be a good thing to extract the root request tracking functionality so we can inject it in Azure Functions.

**Describe the solution you'd like**
This functionality to-be-extracted is about 1) extracting information and 2) tracking that information.
Because 

**Describe alternatives you've considered**

**Additional context**
Originally suggested to be included in the Azure Functions HTTP trigger project template by @pim-simons in the discussion https://github.com/arcus-azure/arcus.templates/discussions/300
Tracked to be included in the HTTP trigger project template https://github.com/arcus-azure/arcus.templates/issues/312

stijnmoreels:
In a first phase, we may consider just copying the request tracking functionality outside the request tracking middleware, to support backwards compatibility and to not directly break anything.

We can mark the old way as obsolete, and provide a fallback with the same functionality as backup until we move towards the next major.

stijnmoreels:
We will also need to implement a dedicated Azure Functions middleware, inheriting from `IFunctionsWorkerMiddleware`.

stijnmoreels:
Probably not that much added-value as the only real functionality that could be exposed in non-out-of-process Functions is the `LogRequest` method itself.

stijnmoreels:
Let's close this as we have done it without: https://github.com/arcus-azure/arcus.templates/pull/615

PR #202, opened 2021-03-02
Feature - skip request tracking attribute

### LINKED ISSUE #200: Provide capability to exclude API endpoints for request tracking
Provide the capability to exclude API endpoints for request tracking since they might not be relevant, or you want to optimize the cost for all your telemetry.

In some cases; you'd like to filter all requests while for others you might want to filter them out based on the status code.

I'd love to have flexibility in the sense that:
- I can apply an attribute on my operation to opt-out for tracking
  - Would be nice to have optional flags there to give granularity to only opt-out for request, response or both
  - Provide an optional flag to define what status codes should be tracked, all that are not specified are considered to be excluded
- Same as above, but through the request logging options instead

That said, I personally, would be fine if we start with attributes only since describing all of this through the options would be either hard or verbose and not really what we would use I think

stijnmoreels:
Adds a `SkipRequestTrackingAttribute` so consumers can pick and choose which endpoints should be tracked and which not.

Relates to #200

### DIFF
--- a/docs/preview/features/logging.md
+++ b/docs/preview/features/logging.md
@@ -78,6 +78,14 @@ public class Startup
 {
     public void Configure(IApplicationBuilder app, IWebHostEvironment env)
     {
+         // In versions less than .NET Core 3.0:
+        // make sure that the endpoint routing is specified before the `UseRequestRouting` if you want to make use of the `SkipRequestTrackingAttribute`.
+        app.UseEndpointRouting();
+
+        // In versions starting from .NET Core 3.0:
+        // make sure that the endpoint routing is specified before the `UseRequestRouting` if you want to make use of the `SkipRequestTrackingAttribute`.
+        app.UseRouting();
+
         app.UseRequestTracking();
 
         ...
@@ -98,6 +106,14 @@ public class Startup
 {
     public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
     {
+        // In versions less than .NET Core 3.0:
+        // make sure that the endpoint routing is specified before the `UseRequestRouting` if you want to make use of the `SkipRequestTrackingAttribute`.
+        app.UseEndpointRouting();
+
+        // In versions starting from .NET Core 3.0:
+        // make sure that the endpoint routing is specified before the `UseRequestRouting` if you want to make use of the `SkipRequestTrackingAttribute`.
+        app.UseRouting();
+
         app.UseRequestTracking(options =>
         {
             // Whether or not the HTTP request body should be included in the request tracking logging emits.
@@ -114,10 +130,32 @@ public class Startup
         });
 
         ...
+
         app.UseMvc();
     }
 }
 ```
+
+**Skipping certain routes**
+
+The request tracking middleware can be disabled for certain endpoints (operation and/or controller level). This can come in handy if you have certain endpoints with rather large request bodies or want to boost performace.
+This can be achieved by placing the `SkipRequestTrackingAttribute` on the endpoint of your choosing.
+
+```csharp
+[ApiController]
+public class OrderController : ControllerBase
+{
+    [HttpPost]
+    [SkipRequestTracking]
+    public IActionResult BigPost()
+    {
+        Stream bigRequestBody = Request.Body;
+        return Ok();
+    }
+}
+```
+
+**Customization**
 
 Optionally, one can inherit from this middleware component and override the default request header sanitization to run some custom functionality during the filtering.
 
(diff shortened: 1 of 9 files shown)


tomkerkhove:
I see you've added a closed tag but can you then also implement the following please?

> - Would be nice to have optional flags there to give granularity to only opt-out for request, response or both
> - Provide an optional flag to define what status codes should be tracked, all that are not specified are considered to be excluded

stijnmoreels:
> I see you've added a closed tag but can you then also implement the following please?
> 
> > * Would be nice to have optional flags there to give granularity to only opt-out for request, response or both
> > * Provide an optional flag to define what status codes should be tracked, all that are not specified are considered to be excluded

Hmm, yeah, will make sure that this PR will not close that issue so we can work on it further. It seemed more workable to implement this in steps since I added also #201 which will take some reworking.

tomkerkhove:
Fine for me, just wanted to make sure we implement it.

docs/preview/features/logging.md
```diff
@@ -114,11 +130,33 @@ public class Startup
         });
 
         ...
+
         app.UseMvc();
     }
 }
 ```
 
+**Skipping certain routes**
```
tomkerkhove, line 139: Just out of curiosity, why are we not doing this?
```suggestion
### Skipping certain routes
```

docs/preview/features/logging.md
```diff
@@ -114,11 +130,33 @@ public class Startup
...
+    [SkipRequestTracking]
+    public IActionResult BigPost()
+    {
+        Stream bigRequestBody = Request.Body;
+        return Ok();
+    }
+}
+```
+
+**Customization**
```
tomkerkhove, line 158: Just out of curiosity, why are we not doing this?
```suggestion
### Customization
```

docs/preview/features/logging.md
```diff
@@ -114,11 +130,33 @@ public class Startup
...
         ...
+
         app.UseMvc();
     }
 }
 ```
 
+**Skipping certain routes**
+
+The request tracking middleware can be disabled for certain endpoints (operation and/or controller level). This can come in handy if you have certain endpoints with rather large request bodies or want to boost performace.
```
tomkerkhove, line 141:
```suggestion
You can opt-out for request tracking on one or more endpoints (operation and/or controller level).

This can come in handy if you have certain endpoints with rather large request bodies or want to boost performance or are constantly probed to monitor the application.
```

docs/preview/features/logging.md
```diff
@@ -114,11 +130,33 @@ public class Startup
...
+This can be achieved by placing the `SkipRequestTrackingAttribute` on the endpoint of your choosing.
```
tomkerkhove, line 142:
```suggestion

This can easily be done by using the `SkipRequestTrackingAttribute` on the endpoint or controller of your choosing.
```

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -46,23 +48,44 @@ public class RequestTrackingMiddleware
...
+            Guard.NotNull(httpContext.Response, nameof(httpContext), "Requires a HTTP response in the context to track the request");
 
-            string requestBody = null;
-            if (_options.IncludeRequestBody)
+            var endpoint = httpContext.Features.Get<IEndpointFeature>();
+            if (endpoint is null)
             {
-                httpContext.Request.EnableBuffering();
-                requestBody = await GetRequestBodyAsync(httpContext);
+                _logger.LogWarning(
```
tomkerkhove, line 58: I'd go with a trace here personally - It's not a security risk & still important to have observability, but not worth the tons of telemetry it will generate
```suggestion
                _logger.LogTrace(
```

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -46,23 +48,44 @@ public class RequestTrackingMiddleware
...
+            
+            var skipRequestTrackingAttribute = endpoint?.Endpoint.Metadata.GetMetadata<SkipRequestTrackingAttribute>();
+            if (skipRequestTrackingAttribute != null)
             {
+                _logger.LogTrace("Skip request tracking for endpoint '{Endpoint}' due to the '{Attribute}' attribute on the endpoint", endpoint.Endpoint.DisplayName, nameof(SkipRequestTrackingAttribute));
                 await _next(httpContext);
             }
-            finally
+            else
             {
```
tomkerkhove, line 70: Maybe move everything in here to a dedicated method?

tomkerkhove (approved):
LGTM - Just a few suggestions

tomkerkhove:
Ok squash merge isn't liking comments when approved

PR #203, opened 2021-03-04
feat: include response body with request tracking

### LINKED ISSUE #199: Provide support for response body tracking
Provide support for response body tracking, in addition to request body tracking.

Allow customers to opt-in through `IncludeResponseBody` which populates a new dimension called `ResponseBody` or so.

When doing this, we should:
- Introduce `RequestBody` which contains what `Body` contains today
- Deprecate `Body`, and add fixed `Request body is now available in 'RequestBody' dimension`
  - Although you could argue that's a breaking change as well, so could just add the request body twice, but that has a cost impact

stijnmoreels:
Adds the option to track also the response body in the request tracking by temporary storing the body and replacing it afterwards. Includes also a maximum tracking option so we don't overload any dependent tracking system (like logging > GB on data).

Closes #199
Relates to #200


src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -54,35 +59,73 @@ public async Task Invoke(HttpContext httpContext)
...
-            finally
+        }
+        
+        private Stream DetermineResponseBodyBuffer()
+        {
+            if (_options.IncludeResponseBody)
             {
-                stopwatch.Stop();
-                TrackRequest(requestBody, httpContext, stopwatch.Elapsed);
+                var responseBodyBuffer = new MemoryStream();
```
stijnmoreels, line 99: Something like a virtual stream that overloads to disk or something may be better than just a memory stream.
Or maybe something like this: https://www.nuget.org/packages/Microsoft.IO.RecyclableMemoryStream/

docs/preview/features/logging.md
```diff
@@ -111,6 +111,14 @@ public class Startup
             // All omitted HTTP request header names that should always be excluded from the request tracking logging emits.
             // (default: `[ "Authentication", "X-Api-Key", "X-ARR-ClientCert" ]`)
             options.OmittedRequestHeaderNames.Add("X-My-Secret-Header");
+
+            // Size of the request body buffer (in bytes) which should be tracked. Request bodies greater than this buffer will only be partly present in the telemetry.
+            // (default: 1MB)
+            options.RequestBodyBufferSize = 10 * 1024 * 1024; // 10MB
```
tomkerkhove, line 117: I presume this is optional?

docs/preview/features/logging.md
```diff
@@ -111,6 +111,14 @@ public class Startup
...
+
+            // Size of the response body buffer (in bytes) hwhich should be tracked. Response bodies greater than this buffer will only be partly present in the telemetry.
+            // (default: 1MB)
+            options.ResponseBodyBufferSize = 10 * 1024 * 1024; // 10MB
```
tomkerkhove, line 121: Same here

docs/preview/features/logging.md
```diff
@@ -111,6 +111,14 @@ public class Startup
             // All omitted HTTP request header names that should always be excluded from the request tracking logging emits.
```
tomkerkhove, line 111: We're missing the opt-in here for the body

tomkerkhove (commented):
Before I go any further, aren't we making it more complex than it is? Do we need to do the size determination and all?

docs/preview/features/logging.md
stijnmoreels: Yes, it's default on 1MB, which means only a MB on data is going to be send to the tracking system (which is already a lot for a single request I think).

tomkerkhove: Question is, why bother?

stijnmoreels:
> Before I go any further, aren't we making it more complex than it is? Do we need to do the size determination and all?

You mean mostly the size determination, right? It's something I chose to combine with this PR as it's very related.
I think with other logging sinks (not everyone is going to use Application Insights solely I presume) we should make sure that we don't overload other systems, right? An malicious user could otherwise with certain endpoints that generate a lot of data push a dependent logging system down. Or am I imagining things here? 😅

tomkerkhove:
Yes. I can see why you are doing it but the default should be include everything. If a size is defined, only then filter out based on that.

Makes sense?

stijnmoreels:
> Yes. I can see why you are doing it but the default should be include everything. If a size is defined, only then filter out based on that.
> 
> Makes sense?

Ah, yes, that will be better maybe. Otherwise it's maybe to 'magic' if users only see a part of their data in the logs.

docs/preview/features/logging.md
stijnmoreels: See above, for same remark-response. Resolve this in favor of that one.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -133,20 +179,38 @@ private async Task<string> GetRequestBodyAsync(HttpContext httpContext)
...
+        {
+            _logger.LogTrace("Prepare for {Target} body to be tracked...", target);
+            string sanitizedBody = await SanitizeStreamAsync(body, maxLength, target);
+            if (sanitizedBody != null)
+            {
+                _logger.LogTrace("Found {Target} body to be tracked", target);
+                return sanitizedBody;
+            }
+                
+            _logger.LogWarning("No {Target} body was found to be tracked", target);
```
tomkerkhove, line 210: I'd use a trace, otherwise for OK/NoContent/Accepted responses we will always generate warnings
```suggestion
            _logger.LogTrace("No {Target} body was found to be tracked", target);
```

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -68,13 +71,28 @@ public async Task Invoke(HttpContext httpContext)
...
-                string requestBody = null;
-                if (_options.IncludeRequestBody)
+            Stream originalResponseBodyStream = null;
+            using (Stream temporaryResponseBodyStream = DetermineResponseBodyBuffer())
+            {
+                if (_options.IncludeResponseBody)
                 {
-                    httpContext.Request.EnableBuffering();
-                    requestBody = await GetRequestBodyAsync(httpContext);
+                    originalResponseBodyStream = httpContext.Response.Body;
```
tomkerkhove, line 94: Why are we doing this exactly? Because we want to limit the amount of tracked content when specified, and we don't want to influence the actual response or?

Just trying to understand, not critisizm.

I'd move this and everything on https://github.com/arcus-azure/arcus.webapi/pull/203/files#diff-3548ef76495796450e2d323886652ae89240370bb37e6ff76b9c4a7565f0e3c5R111-R115 to a dedicated method for better reading. For example https://github.com/arcus-azure/arcus.webapi/pull/203/files#diff-3548ef76495796450e2d323886652ae89240370bb37e6ff76b9c4a7565f0e3c5R109-R115 can be moved to `CopyTemporaryStreamToOriginalResponseStreamAsync`?

Might be good to add a comment why we are doing this as well.

tomkerkhove (approved):

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: Ah, yes, that would be better indeed! Thanks.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: Yeah, just like the request stream, the response stream is not seekable. But unlike the request stream, we don't have a `.EnableBuffering` that makes the body available for multiple reads. So we have to temporary store the response's body in our own seekable stream which we inspect after the processing and track (to app insights or whatever). But after we have tracked the response body (and stream's position is altered), we need to make sure that the remaining middleware components after this one will have the complete response stream. That's why we first reset the position, and then write our temporary stream a second time to the original response stream. 
For the remaining middleware components, it will look like the response was created for the first time.

Makes somewhat sense?
I'll check the links you send to see if we can indeed move something to a separate method, but I would be careful with this since it can make things more complicated.

tomkerkhove: Certainly, thanks for sharing. I'd add this briefly as a comment, personally.

stijnmoreels: Ok, I'll describe the comment there more for future references.

stijnmoreels:
I'll merge this, ok?

PR #211, opened 2021-03-30
feat: attribute to exclude request/response bodies from request tracking

stijnmoreels:
Provide the capability to decorate certain routes with the `ExcludeRequestTrackingAttribute` which can control more fine-grained exclusion filters on the request/response bodies.

Relates to #200


docs/preview/features/logging.md
```diff
@@ -168,6 +168,37 @@ public class OrderController : ControllerBase
 }
 ```
 
+When used as in the example above, the all routes of the controller will be excluded from the telemtry tracking.
```
fgheysels, line 171:
```suggestion
When used as in the example above, then all routes of the controller will be excluded from the telemetry tracking. 
```

docs/preview/features/logging.md
```diff
@@ -168,6 +168,37 @@ public class OrderController : ControllerBase
...
+The exclude attribute allows you to specify on a more specific level what part of the request should be excluded.
+
+```csharp
+[ApiController]
+public class OrderController : ControllerBase
+{
+    // Only exclude the request body from the request tracking. 
+    // The request will still be tracked and will contain the request headers and the response body (if configured).
+    [HttpPost]
+    [ExcludeRequstTracking(ExcludeFilter.ExcludeRequestBody)]
```
fgheysels, line 181:
```suggestion
    [ExcludeRequestTracking(ExcludeFilter.ExcludeRequestBody)]
```

fgheysels, line 181: I need to sleep over this :)
Isn't it a bit confusing to have a double negation ?
I mean, we assign the attribute `ExcludeRequestTracking` which would mean that we do not track any requests.  But then, what does the `ExcludeFilter` mean ?  Then you say:  hey, do request tracking, but don't include the body.
I think it's a bit confusing when reading the code , since you're actually doing request-tracking.

Why don't we do it like this:

```
[RequestTracking(Options.Disable)]    // Disables all request tracking
[RequestTracking(Options.ExcludeRequestBody)]  // Request tracking, but do not include the body
....
```

Just a thought.

stijnmoreels: Hmm yeah it's more like either use the `[ExcludeRequestTracking]` to exclude everything (meaning the entire route will be skipped), or use the `ExcludeFilter` to filter out only parts and not everything.
So, if I'm reading it right you're seeing it the way it's proposed here. It seems logical to have 'exclude' in the name of the attribute bc in every possible scenario it will exclude something. Maybe the name of the filter should be renamed then...

stijnmoreels: > Does the ExcludeRequestTracking attribute now prevent that any dependency is tracked, or not ? 

Yes, it does. When you only use the `[ExcludeRequestTracking]` on either a controller or a method, the request tracking will be excluded completely. No request telemetry will be written altogether.

> Or is it just a way to have some more control over what is being tracked ?

Using the filter, you can only exclude some parts of the request tracking instead of the default, everything.

fgheysels: So, when you decorate a controller or method like this:

```csharp
[ExcludeRequestTracking(Options.ExcludeRequestBody)]
public IActionResult Foo(){}
```
Request tracking will be enabled, but the body will not be written.   So, imho, it's a bit counter-intuitive, as the ExcludeRequestTracking attribute is applied, and I would suppose that it doesn't do any request-tracking at all.

But maybe that 's just me .... 

Does this attribute work in conjunction with some other middleware ?

stijnmoreels: Yes, on your first question. It was the way we agreed upon in #200 . But maybe we should then work with separate attributes, each for one option, then? Like `ExcludeRequestBodyInRequestTracking` or so? With the options now, it seemed like a good choice bc you could use one attribute with OR'ed enum flags to customize it. (Like `Exclude.RequestBody | Exclude.ResponseBody`).

For you second question, no it doesn't directly work with other middleware. But the attribute introduced here can be applied multiple times and the request tracking middleware will merge the specified options into a final exclusion result.

fgheysels: or maybe just rename the attribute and make the Options mandatory ?

Like this:

``` 
[RequestTrackingOptions(Options.ExcldueRequestTracking)]
[RequestTradckingOptions(Options.ExcludeBody)]
[RequestTrackingOptions(Options.ExcludeHeaders)]
[RequestTrackingOptions(Options.ExcludeBody | Options;ExcludeHeaders)]
```

src/Arcus.WebApi.Logging/SkipRequestTrackingAttribute.cs
```diff
@@ -1,11 +0,0 @@
-﻿using System;
-
-namespace Arcus.WebApi.Logging
-{
-    /// <summary>
-    /// Represents an endpoint attribute that indicates which endpoints should be withhold from request tracking.
-    /// </summary>
-    public class SkipRequestTrackingAttribute : Attribute
```
tomkerkhove, line 8: Careful here, we'll need to be backwards compatible on this one

tomkerkhove (approved):
LGTM, just one remark

src/Arcus.WebApi.Logging/SkipRequestTrackingAttribute.cs
stijnmoreels: No need, I double checked it and we didn't released this attribute yet, so we're free to change it. 👍

docs/preview/features/logging.md
stijnmoreels: What do you think @tomkerkhove , something like this?

stijnmoreels: I think it's a valuable request, only when it's gets used to exclude the entire route, it may look a bit strange:

```csharp
[RequestTrackingOptions(Options.ExcludeRoute)]
```

fgheysels: The `Options` name was merely an example.  As you know, naming is one of the hardest parts of our job :) so I didn't spend too much time of finding an appropriate name :)
In the example however, I'm suggesting however to make the `Options` enum (ok, a better name must be found) to be a flags-enum.

fgheysels: > I think it's a valuable request, only when it's gets used to exclude the entire route, it may look a bit strange:
> 
> ```cs
> [RequestTrackingOptions(Options.ExcludeRoute)]
> ```
How's that ?  Maybe I'm missing something, but I don't see it ?

stijnmoreels: Hmm, well, it may be strange for users to have this once next to the other options (exlude request body, response body..) as it does that for them. The flagged enum I use here is showing that in the integer value (7) that it includes all the previously defined options; and by having a `[ExcludeRequestTracking]` without any args simplifies this enormously (which behind the scenes uses this `ExcludeRoute` enum member).

In the proposed solution, we'll have to explicitly specify this, maybe making it a bit more pain to use.

stijnmoreels: Maybe, let's wait for Tom, he can maybe help with this 🥇 .

tomkerkhove: Or use 2 attributes? :)

- `ExcludeRequestTracking` removes everything
- `RequestTracking(Exclude.RequestBody)` removes a subset

stijnmoreels: @fgheysels , you can agree with this?

fgheysels: I can agree with it.
But:  Can you then also add `RequestTrackingAttribute` without any options, or not ?

stijnmoreels: No, that would not be the idea here I guess. It's actually a combination of both approaches. A single attribute without options to exclude routes and a separate one with (required) options.
Seems ok to you?☺️

stijnmoreels: I'll assume that this is okay for you. I'll continue with resolving and merging but don't hesitate to create an issue if something's still not ok for you.

PR #214, opened 2021-04-07
feat: Authorization header should not be tracked

tomkerkhove:


stijnmoreels (approved):
Yes, was thinking about that too. Thanks!

PR #215, opened 2021-04-09
feat: tracked response statuscodes

### LINKED ISSUE #200: Provide capability to exclude API endpoints for request tracking

stijnmoreels:
Provides an additional request tracking option that describes what response status codes should be tracked.
Provides an additional input for the request tracking options attribute that describes the same thing.

Closes #200

### DIFF
--- a/docs/preview/features/logging.md
+++ b/docs/preview/features/logging.md
@@ -41,15 +41,15 @@ public class Startup
 {
     public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
     {
-       app.UseExceptionHandling();
+        app.UseExceptionHandling();
 
-       ...
-       app.UseMvc();
+        ...
+        app.UseMvc();
     }
 }
 ```
 
-If you have multiple middleware components configured, it is advised to put the `ExceptionHandlingMiddleware` as the first one.
+If you have multiple middleware components configured, it is advised to put the `ExceptionHandlingMiddleware` as soon as possible.
 By doing so, unhandled exceptions that might occur in other middleware components will also be logged by the `ExceptionHandlingMiddleware` component.
 
 ## Logging incoming requests
@@ -88,8 +88,10 @@ public class Startup
 
         app.UseRequestTracking();
 
-        ...
+        // Make sure that the exception handling is placed after the request tracking because otherwise unhandled exceptions that result in 500 response status codes will not be tracked.
+        app.UseExceptionHandling();
         app.UseMvc();
+        ...
     }
 }
 ```
@@ -139,6 +141,10 @@ public class Startup
             // Size of the response body buffer (in bytes) hwhich should be tracked. Response bodies greater than this buffer will only be partly present in the telemetry.
             // (default: no limit)
             options.ResponseBodyBufferSize = 10 * 1024 * 1024; // 10MB
+
+            // All omitted HTTP routes that should be excluded from the request tracking logging emits.
+            // (default: no routes).
+            options.OmittedRoutes.Add("/api/v1/health");
         });
 
         ...
--- a/docs/preview/features/logging.md
+++ b/docs/preview/features/logging.md
@@ -168,6 +168,8 @@ public class OrderController : ControllerBase
 }
 ```
 
+#### Excluding request/response bodies
+
 When used as in the example above, then all routes of the controller will be excluded from the telemetry tracking. 
 The exclude attribute allows you to specify on a more specific level what part of the request should be excluded.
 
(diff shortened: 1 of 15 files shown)


tomkerkhove:
In this case we are more using an opt-in instead of an opt-out, no? Ideally I'd be able to say "don't track 200 OK, but do track all the rest!"

stijnmoreels:
> In this case we are more using an opt-in instead of an opt-out, no? Ideally I'd be able to say "don't track 200 OK, but do track all the rest!"

Yes, this is an opt-in as described in the issue: _"Provide an optional flag to define what status codes should be tracked, all that are not specified are considered to be excluded"_.

tomkerkhove:
Hm, I've misformulated it then but that's ok.

tomkerkhove:
One thing I wwould add then is support for ranges so that I can include everything in the 500 or 500-505 range.

The reason why we need this, for example, is to reduce the telemetry that our health endpoint is generating so that we only have the 5xx responses.

docs/preview/features/logging.md
```diff
@@ -255,6 +257,75 @@ So the resulting log message becomes:
...
+
+```csharp
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+
+public class Startup
+{
+    public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+    {
+        app.UseRequestTracking(options => options.TrackedStatusCodes.Add(HttpStatusCode.Created));
```
tomkerkhove, line 296: We are using Dapr which add a probe endpoint that we don't control.

In order to filter that out, we would maybe need to define this here somehow based on the path.
How hard would it be to add that?

stijnmoreels: That's more something for another issue and PR, but I'm thinking about routes or route ranges (wildcards) that you could add to the options if you can't control it yourself.
In other scenario's, you could use our `ExcludeRequestTracking` attribute I guess.

tomkerkhove: It's not really the same as we want to limit on a set of status codes, not request/response payload.

tomkerkhove: Seperate task for this is ok though.

docs/preview/features/logging.md
```diff
@@ -255,6 +257,75 @@ So the resulting log message becomes:
 
 `HTTP Request POST http://localhost:5000/weatherforecast completed with 200 in 00:00:00.0191554 at 03/23/2020 10:12:55 +00:00 - (Context: [X-Api-Key,])`
 
+#### Tracked HTTP status codes
```
tomkerkhove, line 260:
```suggestion
#### Reducing requests to specific HTTP status code(s)
```

docs/preview/features/logging.md
```diff
@@ -255,6 +257,75 @@ So the resulting log message becomes:
...
+More grainer control can be achieved via placing an attribute on either the controller's class definition or the endpoint method:
+
+```csharp
+using Arcus.WebApi.Logging;
+
+[ApiController]
+public class OrderController : ControllerBase
+{
+    [HttpPost]
+    [RequestTracking(HttpStatusCode.Created)]
```
tomkerkhove, line 311: Would love to support this:
```suggestion
    [RequestTracking(HttpStatusCode.Created, HttpStatusCode.OK)]
```

Or define a range from 200 -> 299.

docs/preview/features/logging.md
```diff
@@ -168,6 +168,8 @@ public class OrderController : ControllerBase
 }
 ```
 
+#### Excluding request/response bodies
```
tomkerkhove, line 171: Do we need to update a TOC?

docs/preview/features/logging.md
stijnmoreels: Ah, that would be better. Now it's supported for multiple ones that you can inside one attribute but we could use it indeed for ranges. That would be better maybe.

stijnmoreels: But let's do that in another issue too. Now it already supports this syntax but it takes the two status codes separatly in. Maybe for ranges, we should come up with our own enumeration or let it specify as integers.

docs/preview/features/logging.md
stijnmoreels: No, as the TOC only contains the main headers.

docs/preview/features/logging.md
stijnmoreels: I'll remove the `params` definition for now so we can add these ranges later.

stijnmoreels: Ok, removed these `params`. I'll create an issue for the ranges.

stijnmoreels: Done https://github.com/arcus-azure/arcus.webapi/issues/216, resolving.

docs/preview/features/logging.md
stijnmoreels: Ok, not entirely sure what the goal is here, so I suggest that you make the additional issue for this one. 😄 I'll resolve this comment.

tomkerkhove (approved):

PR #218, opened 2021-04-23
feat: omit routes in request tracking

### LINKED ISSUE #217: Provide capability to exclude request tracking for a URI
Provide capability to exclude request tracking for a URI for which you do not own the API endpoint.

This means that we are not able to add an attribute on the operation, but instead specify something like `/healthz` to be excluded.

Be able to define one or more URIs to exclude during the registration of the request tracking as part of the options.

**Describe alternatives you've considered**
A Serilog filter can be written to exclude this, but ideally, we can manage it as part of the middleware.

**Additional context**
This is required for cases where tools like Dapr automatically add endpoints under the hood:
https://github.com/dapr/dotnet-sdk/blob/fdc35bbbafbffe4ad85663441453c5aec343244a/src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs#L134-L147:
```csharp
        private static IEndpointConventionBuilder MapActorHealthChecks(this IEndpointRouteBuilder endpoints)
        {
            var builder = endpoints.MapHealthChecks("/healthz");
            builder.Add(b => 
            {
                // Sets the route order so that this is matched with lower priority than an endpoint
                // configured by default.
                //
                // This is necessary because it allows a user defined `/healthz` endpoint to win in the
                // most common cases, but still fulfills Dapr's contract when the user doesn't have
                // a health check at `/healthz`.
                ((RouteEndpointBuilder)b).Order = 100;
            });
            return builder.WithDisplayName(b => "Dapr Actors Health Check");
```

stijnmoreels:
Make sure that we can exclude entire HTTP routes from the request tracking options instead of using attributes.

Closes #217

### DIFF
--- a/docs/preview/features/logging.md
+++ b/docs/preview/features/logging.md
@@ -139,6 +139,10 @@ public class Startup
             // Size of the response body buffer (in bytes) hwhich should be tracked. Response bodies greater than this buffer will only be partly present in the telemetry.
             // (default: no limit)
             options.ResponseBodyBufferSize = 10 * 1024 * 1024; // 10MB
+
+            // All omitted HTTP routes that should be excluded from the request tracking logging emits.
+            // (default: no routes).
+            options.OmittedRoutes.Add("/api/v1/health");
         });
 
         ...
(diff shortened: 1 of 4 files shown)


tomkerkhove (approved):
Nice and clean 👌

Issue #216, opened 2021-04-16
Provide support to include a range of HTTP status codes to be tracked


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
In a previous feature enhancement (#215) we introduced single HTTP status codes from being included in the request tracking. When supporting a whole range of HTTP status codes, it may be cumbersome to include them all manually.

**Describe the solution you'd like**
It would be good that the `[RequestTracking]` attribute allows us to set a range of HTTP status codes. To-be-determined if this has to be done with integers and/or the built-in enum or even a custom enumeration of our own.

**Additional context**
Continuation of #200

PR #220, opened 2021-04-27
fix: add extra null-check for HTTP status code range

stijnmoreels:
Add extra `null`-check before we check if the response status code is within the configured range.

Relates to #216


tomkerkhove (approved):
Can we add a test for this?

stijnmoreels:
> Can we add a test for this?

Yes, will do! 👍

PR #224, opened 2021-04-29
fix: Formatting for shared access key is broken

tomkerkhove:
Formatting for shared access key is broken in rendered result


fgheysels (approved):
Increasing readability is always a good thing.
There's no such space as whitespace.

PR #229, opened 2021-05-04
docs: UseRequestTracking instead of UseRequestRouting

fgheysels:
Minor correction in the documentation:
`UseRequestRouting` was mentionned instead of `UseRequestTracking` in the comments on how to use this feature.


tomkerkhove (approved):
Thank you!

Issue #230, opened 2021-05-04
UseRequestTracking logs an additional request with a wrong timestamp


fgheysels:
**Describe the bug**
I'm using Arcus.WebApi.Logging v1.3.1 and I've enabled `UseRequestTracking`:

```csharp
app.UseHttpCorrelation();
app.UseRouting();
app.UseRequestTracking();
app.UseAuthentication();
```

I'm also logging to Application Insights.

After enabling `UseREquestTracking`  I can see that an additional request is logged in AppInsights:

![image](https://user-images.githubusercontent.com/3605786/117042454-4b54e600-ad0c-11eb-8735-cc33b58db19c.png)

In the above screenshot, you can also see that the timestamp of that request is wrong: it is not precise enough.

The precision of the timestamp is only set to seconds, while it should be set to the tenth millions of a second (`yyyy-MM-ddTHH:mm:ss.fffffff zzz`).  (See also PR 181 in Arcus.Observability).

![image](https://user-images.githubusercontent.com/3605786/117042518-6162a680-ad0c-11eb-8e04-1090c29bb962.png)

tomkerkhove:
Relates to https://github.com/arcus-azure/arcus.observability/pull/181

tomkerkhove:
This is because we are still using v2.0.0 apparantly:
https://github.com/arcus-azure/arcus.webapi/blob/master/src/Arcus.WebApi.Logging/Arcus.WebApi.Logging.csproj#L33-L36

tomkerkhove:
See https://github.com/arcus-azure/arcus/issues/129

fgheysels:
Can this one be closed ?

tomkerkhove:
I think we can but will let @stijnmoreels verify

PR #232, opened 2021-05-06
fix: use Arcus NuGet version ranges

stijnmoreels:
Let's use NuGet version ranges to include Arcus dependencies so we can more easily support multiple Arcus packages in a single project.

Relates to https://github.com/arcus-azure/arcus/issues/129

### DIFF
--- a/src/Arcus.WebApi.Logging.Core/Arcus.WebApi.Logging.Core.csproj
+++ b/src/Arcus.WebApi.Logging.Core/Arcus.WebApi.Logging.Core.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


stijnmoreels:
Something like this, @tomkerkhove ?

tomkerkhove:
No because now you're using 3.0 as inclusive but it should be exclusive

stijnmoreels:
> No because now you're using 3.0 as inclusive but it should be exclusive

Just saw it, pushed already. 😄

stijnmoreels:
> No because now you're using 3.0 as inclusive but it should be exclusive

Just saw it, pushed already. 😄

tomkerkhove (approved):

tomkerkhove:
Based on https://github.com/arcus-azure/arcus.webapi/issues/230 I would actually increase it to at minimum `2.0.1`

PR #234, opened 2021-05-13
feat: opt-in for security events

### LINKED ISSUE #225: Provide capability to opt-in for security events
Provide capability to opt-in for security events so that consumers can choose if they want to have them or not.

Having an opt-in would be nice since it tends to generate a lot of telemetry if it's turned on by default.

**Additional context**
We are currently seeing a high volume of security events while we are not interested in them for the time being.

image

stijnmoreels:
Provide the capability to opt-in for security events in the authentication and authorization components.

Closes #225


src/Arcus.WebApi.Security/Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
```diff
@@ -25,14 +26,27 @@ public class SharedAccessKeyAuthenticationAttribute : TypeFilterAttribute
...
+            const bool emitSecurityEvents = false;
+            Arguments = new object[] { headerName?? String.Empty, queryParameterName?? String.Empty, secretName, emitSecurityEvents };
+        }
+
+        /// <summary>
+        /// Gets or sets the flag indicating whether or not the shared access key authentication should emit security events during the authentication process of the request.
+        /// </summary>
+        public bool EmitSecurityEvents
+        {
+            get => Arguments?.LastOrDefault() is bool emitSecurityEvents && emitSecurityEvents;
```
tomkerkhove, line 48: Cna't we use an object here? What if we add a new argument, then this will be broken, no?

tomkerkhove (approved):

src/Arcus.WebApi.Security/Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: Yes, but that was already the case before these changes. The security attributes already are fixed with their security filter counterparts. The attributes reference towards the type of the filter and send a series of arguments with it to create an filter instance. So, it would already break but I guess in this case it's okay because they 'work together'. They require each other.
Adding this extra property to alter the arguments as an 'afterthought' doesn't make it weaker. Our tests will make sure that things still keeps working if we do change things.

stijnmoreels: > Cna't we use an object here? What if we add a new argument, then this will be broken, no?

Not sure what you mean with an 'object', though? We can't alter the filter (which gets created based on the input from the attribute) with a property. So, we have to pass it as an argument.

stijnmoreels:
I'll probably update this PR to take in a set of options instead of a primitive boolean.

src/Arcus.WebApi.Security/Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs
stijnmoreels: After consideration, I removed this direct approach and used a separate options instance that will be altered via the property. Better and safer. Thanks.

docs/preview/features/security/auth/certificate.md
```diff
@@ -56,18 +56,33 @@ public class Startup
...
+
+            // Additional consumer-configurable options to change the behavior of the authentication filter.
+            var authOptions = new CertificateAuthenticationOptions
+            {
+                // Adds certificate authentication to the request pipeline with emitting security events during the authorization of the request.
+                // (default: `false`)
+                EmitSecurityEvents = true;
+            };
+
+            mvcOptions.Filters.Add(new CertificateAuthenticationFilter(authOptions));
```
stijnmoreels, line 84: This makes me think whether we should add an additional extension on the `IFilterCollection` that lets the consumer configure the options inside a lambda.

stijnmoreels: Added these extensions. Question though: these are (like the JWT extension) in the `Microsoft.AspNetCore.Mvc.Filters` namespace because the `FilterCollection` is in that namespace; but this is probably (always?) configured during startup with the `IServiceCollection` services. So, if a consumer uses `services.AddMvc(options => options.Filters.[our extension])` it will now show.
Should we move these extensions then to the `Microsoft.Extensions.DependencyInjection`, which is the same namespace as the `IServiceCollection`?
We can for the JWT extension create a new extension set and obsolete the one we have now if we go through with this.

tomkerkhove (approved):

docs/preview/features/security/auth/certificate.md
tomkerkhove: I think the current one is best since you'll do it on the filters directly, no?

stijnmoreels: Ok, let's leave it then!

tomkerkhove:
@stijnmoreels Are we safe to merge?

stijnmoreels:
> @stijnmoreels Are we safe to merge?

Yes, only I was waiting for @fgheysels , but we can merge otherwise.

PR #275, opened 2021-12-30
docs: move telemetry enrichment

### LINKED ISSUE #273: Remove 'telemetry enrichment' page as it's a duplicate of the 'Serilog logging' section
We maintain a [telemetry enrichment](https://webapi.arcus-azure.net/features/telemetry) feature doc page that is actually a duplicate of the [correlation Serilog section](https://webapi.arcus-azure.net/features/correlation#logging).

We should consolidate it all in the correlation page and update the code samples accordingly (as they are not fully up-to-date). 

Also, we should add a redirect so that the old telemetry page is redirected to this new section (to-be-continued as the frontend team will take this redirection functionality up).

stijnmoreels:
The `telemetry.md` was a duplicate of the section in the `correlation.md` feature docs.
Moved any relevant information to the `correlation.md`, removed the `telemetry.md`, and added a redirect to point to the correlation correlation header.

Closes #273


netlify.toml
```diff
@@ -34,3 +34,7 @@ ID = "b0838013-ab4a-4e8d-a266-0eb61405ad77"
 # inherit these settings.
 # [context.deploy-preview]
   # publish = "dist/"
+
+[[redirects]]
+  from = "/:version/features/telemetry"
+  to = "/:version/features/correlation#logging"
```
fgheysels, line 40: Great that you though of this.

docs/preview/features/correlation.md
```diff
@@ -148,7 +148,16 @@ public class OrderController : ControllerBase
...
+As an additional feature, we provide an extension to use the HTTP correlation directly in a [Serilog](https://serilog.net/) configuration as an [enricher](https://github.com/serilog/serilog/wiki/Enrichment). 
+This adds the correlation information of the current request to the log event as a log property called `TransactionId` and `OperationId`.
+
+**Example**
+
+- `TransactionId`: `A5E90591-ADB0-4A56-818A-AC5C02FBFF5F`
+- `OperationId`: `79BB196A-B0CC-4F5C-B48A-AB87850346AF`
+
+**Usage**
+The enricher requires access to the application services so it can get the correlation information.
```
fgheysels, line 160: Great to have the docs centralized.  I'm just wondering if we should mention here something regarding that `CorrelationInfoEnricher`, or is that now something that is obsolete ?

stijnmoreels: It's not obsolete, but we added an extension for better dev experience. So, the 'old' way is still valid, but not recommended via the docs. We could maybe say something like: ... adds a `CorrelationInfoEnricher` to the Serilog enrichment... or something like that. But not sure if this would provide much added-value, as we should maybe guide people to the extension described here.

netlify.toml
stijnmoreels: Thx! It's a bit tricky though, bc we can't remove the `# Logging` header, otherwise we again have death links 😄 .

fgheysels: But I'd assume we do not need to keep this redirect around for eternity ?

docs/preview/features/correlation.md
fgheysels: I didn't get that from just reviewing :)

fgheysels (approved):

netlify.toml
stijnmoreels: Not sure, 🤷‍♂️.

Issue #250, opened 2021-08-30
Provide extensions for `MvcOptions`


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Provide extensions for `MvcOptions` which are being copied in every Web API template that we offer:
```csharp
private static void RestrictToJsonContentType(MvcOptions options)
{
    var allButJsonInputFormatters = options.InputFormatters.Where(formatter => !(formatter is SystemTextJsonInputFormatter));
    foreach (IInputFormatter inputFormatter in allButJsonInputFormatters)
    {
        options.InputFormatters.Remove(inputFormatter);
    }

    // Removing for text/plain, see https://docs.microsoft.com/en-us/aspnet/core/web-api/advanced/formatting?view=aspnetcore-3.0#special-case-formatters
    options.OutputFormatters.RemoveType<StringOutputFormatter>();
}

private static void ConfigureJsonFormatters(MvcOptions options)
{
    var onlyJsonInputFormatters = options.InputFormatters.OfType<SystemTextJsonInputFormatter>();
    foreach (SystemTextJsonInputFormatter inputFormatter in onlyJsonInputFormatters)
    {
        inputFormatter.SerializerOptions.IgnoreNullValues = true;
        inputFormatter.SerializerOptions.Converters.Add(new JsonStringEnumConverter());
    }

    var onlyJsonOutputFormatters = options.OutputFormatters.OfType<SystemTextJsonOutputFormatter>();
    foreach (SystemTextJsonOutputFormatter outputFormatter in onlyJsonOutputFormatters)
    {
        outputFormatter.SerializerOptions.IgnoreNullValues = true;
        outputFormatter.SerializerOptions.Converters.Add(new JsonStringEnumConverter());
    }
}
```

**Describe the solution you'd like**
Extensions that can be updated regardless of when the template was created.

stijnmoreels:
So, you mean, it's easier to alter an extension in a dependent package then to change it in the template itself?

tomkerkhove:
No, once the template has scaffolded the project we can no longer help end-users. If we move it into a package, we can still make it better and allow them to upgrade.

stijnmoreels:
Hmm, in a separate package? `Arcus.WebApi.Formatting`?

tomkerkhove:
I don't want to create too narrow packages so what about `Arcus.WebApi.Hosting` which has a `Formatting` folder?

What do you think @fgheysels?

PR #324, opened 2022-05-13
feat: add custom healthreport json dto

### LINKED ISSUE #277: Introduce a new `HealthReport` type to use across Arcus repo's
When exposing the Swagger (UI) of your application with the built-in Microsoft [`HeathReport` ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.diagnostics.healthchecks.healthreport?view=dotnet-plat-ext-6.0), you get a whole lot of assembly information because the `Exception` is part of that type. This is not only unnecessary, but could expose a security risk for leaking incontrollable information.

Create a subset of the Microsoft `HealthReport` type with information that we control and are safe to expose. This type should be easily used with the Microsoft variant, and should be serializable easily. We can then use it in our Templates Web API project template and in the TCP probe in Messaging.

**Additional context**
Related to https://github.com/arcus-azure/arcus.templates/issues/491

stijnmoreels:
Creates a dedicated `HealthReportJson` without any exception details as alternative for Microsoft's `HealthReport` so it can be safely used within OpenAPI documents and serialization.

Closes #277

### DIFF
--- a/docs/preview/features/openapi/health-checks.md
+++ b/docs/preview/features/openapi/health-checks.md
@@ -1,0 +1,58 @@
+﻿---
+title: "Using health checks report in OpenAPI documentation"
+layout: default
+---
+
+# Using health checks report in OpenAPI documentation
+
+Microsoft has a way of providing the health status of an application by running over a series of 'health checks' that eventually generate a health checks report. 
+For more information on application health, see [Microsoft's documentation](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks).
+
+The problem is that the health checks report model called `HealthReport` cannot be safely used as a JSON serializable type, as it could contain exception details (and therefore assembly information).
+Arcus provides a dedicated JSON data-trasfer object (DTO) to expose the health status of the application in a reliable and safe manner.
+
+## Installation
+
+This feature requires to install our NuGet package
+
+```shell
+PM > Install-Package Arcus.WebApi.OpenApi.Extensions
+```
+
+## Usage
+
+The Arcus-provided health checks status model is called `HealthReportJson` and has the same information as Microsoft's model but without any exception details.
+The Arcus model and Microsoft's model are perfectly exchangable. They can be converted between each other, but keep in mind that the exception details are not part of this conversion.
+
+Here's an example of an API controller that uses the Microsoft's health checks service to generate the health report, but uses Arcus' model to expose this information.
+Notice that the `[ProduceResponseType]` attribute uses Arcus' model so that the OpenAPI document uses this model during its generation.
+
+```csharp
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+[ApiController]
+[Route("api/v1/health")]
+public class HealthController : ControllerBase
+{
+    private readonly HealthCheckService _healthService;
+
+    public HealthController(HealthCheckService healthService)
+    {
+        _healthService = healthService;
+    }
+
+    [HttpGet]
+    [ProducesResponseType(typeof(HealthReportJson), 200)]
+    public async Task<IActionResult> Get()
+    {
+        HealthReport report = await _healthService.CheckHealthAsync();
+
+        var json = HealthReportJson.FromHealthReport(report);
+        return Ok(json);
+    }
+}
+```
+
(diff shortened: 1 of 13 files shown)


docs/preview/features/openapi/health-checks.md
```diff
@@ -0,0 +1,58 @@
...
+
+The problem is that the health checks report model called `HealthReport` cannot be safely used as a JSON serializable type, as it could contain exception details (and therefore assembly information).
+Arcus provides a dedicated JSON data-trasfer object (DTO) to expose the health status of the application in a reliable and safe manner.
+
+## Installation
+
+This feature requires to install our NuGet package
+
+```shell
+PM > Install-Package Arcus.WebApi.OpenApi.Extensions
```
fgheysels, line 19: I would 've added this to the Arcus.WebApi project itself, instead of adding a separate package.
IMHO, this is not only linked with the OpenAPI docs, it is instead actually the API model of the Health endpoint.

stijnmoreels: It's not added to a new package, though, since we already have something for OpenAPI and since it was OpenAPI that was the cause that this problem bubbled up.
But see your point. What other place could we put it that isn't a new package? Maybe `Arcus.WebApi.Hosting`?

fgheysels: I had to open Arcus.WebApi to see what we all had.   To be honest, I don't find a suitable place right away.

Maybe we should just add it as a class in the WebApi template project ?  After all, it is only there that it is being used.
When people start a new project based on the Arcus Template, they get a predefined health-endpoint.  The HealthReport API model belongs to that endpoint, so I would keep it close to where it is being used atm.
If it turns out later that we can put that class somewhere else, we can always change it.

stijnmoreels: The `HealthReport` is actually also used in the TCP health probe in the messaging projects, so it would be a good thing that it's added there too.

stijnmoreels: Yeah, we could. Was just wondering if this is maybe a bit too low-level. Not very user-friendly to just put it there. It's also one of the reasons we added MVC extensions for the JSON formatting, bc we didn't want too much additional code there for things that will always be the same kind of structure.

fgheysels: I'm not sure what you mean here.

IMHO, the HealthReport class is a (api) model.  It belongs close to the API operation that uses / returns this model.  Therefore, I see no problem in putting it in the Arcus WebAPI template f.i.

stijnmoreels: > I'm not sure what you mean here.

Recently, we updated the Web API template, bc it contained a lot of custom code (that was repeated in each new project creation from the template). This was a lot of duplication and made the overall user experience worse. We decided to extract the custom code to MVC extensions so that we can maintain the custom code better and the user experience is greatly improved. See this issue for more info: https://github.com/arcus-azure/arcus.webapi/issues/250
The reason why I mention this, is that we are now considering to again place custom code in three project templates and while the custom code of the MVC options was mostly configuration code, is the custom health report for serialization an even more 'internal' type. So, bit reluctant to add this just as-is to the template projects as it conflicts with the 'mindset' we were starting with the templates.

stijnmoreels: Maybe, we could include the health report JSON dto to the `Arcus.Messaging.Health` package, and use it there. (not sure if we would qualify it as a breaking change there), as the API operation we set up is set up there.

fgheysels: Is it really custom code ?  It is an API model and imo API models should live close to where they're used.
I don't think we should share the same model between multiple services, as this would mean that they cannot evolve separately.  
The HealthReport model is not infrastructure code, so to me it doesn't feel problematic to add it directly into the template.  Moreover, you allow users of the template to fine-tune / modify that model if they want to.

fgheysels: What exactly do you mean with 'adding it explicitly' ?

stijnmoreels: Adding it to the template, like proposed in: https://github.com/arcus-azure/arcus.templates/pull/589

stijnmoreels: Added it here in the messaging library: https://github.com/arcus-azure/arcus.messaging/issues/287 we can close this PR.

PR #336, opened 2022-06-09
chore: rethink http correlation extension

stijnmoreels:
Proposal to make sure that we don't use deprecated extensions, but don't release a next major, too.

Relates to https://github.com/arcus-azure/arcus.templates/issues/601


stijnmoreels:
A little help?
This is technically a breaking change, so shouldn't be done just yet for the v1.6 release, BUT, the way it is changed wouldn't cause any major problems. The only thing I can think of is that people would get reflection issues because the signature has changed.

fgheysels:
> A little help? This is technically a breaking change, so shouldn't be done just yet for the v1.6 release, BUT, the way it is changed wouldn't cause any major problems. The only thing I can think of is that people would get reflection issues because the signature has changed.

I think results from the survey have shown that ppl want us to release faster, so maybe we should not be afraid of breaking changes.  I think changes like this one can be easily fixed when ppl upgrade ?

fgheysels (approved):

PR #342, opened 2022-06-20
feat: open exception handling middelware for extension

### LINKED ISSUE #340: Open ExceptionHandlingMiddleware for extension
We have our ExceptionHandlingMiddleware component which logs unhandled exceptions to AppInsights and return a http 500 statuscode.

I think it would be good if users can specify what should be logged to application insights, and what should be returned by the middleware to the caller.

An idea to do that, would be to make the `LogException` method `virtual protected` instead of `private`.
We could do something similar for the response that is being returned; move that code to a `virtual protected` method.

Main question would be:  how does this fit in with the helper method which registers the middleware in the pipeline.   I see that a developer might want to inherit from the Arcus ExceptionHandlingMIddleware, but then, he needs to register his custom middleware component that inherits from the Arcus middleware.

Would we be able to do that via 

```csharp
app.UseExceptionHandling<MyExceptionHandlingMiddleware>();
```

That would mean that we need to provide an extension method that takes a type parameter, with the constraint that the type-parameter must inherit from the Arcus ExceptionHandlingMiddleware component.

stijnmoreels:
Open the `ExceptionHandlingMiddleware` component for extension.

Closes #340

### DIFF
--- a/docs/preview/features/logging.md
+++ b/docs/preview/features/logging.md
@@ -65,7 +65,7 @@ By doing so, unhandled exceptions that might occur in other middleware component
 
 When custom exception handling is required, you can inherit from the `ExceptionHandlingMiddleware` to create your own middleware component and register it with Arcus' extension.
 
-This example implements the exception handling middleware to influence the log message and returns a custom determination of the HTTP response status code.
+This example implements the exception handling middleware to influence the log message and adds a custom determination of the HTTP response status code.
 
 ```csharp
 using Arcus.WebApi.Logging;
@@ -78,25 +78,25 @@ public class MyExceptionHandlingMiddleware : ExceptionHandlingMiddleware
     {
     }
 
-    protected override void LogException(ILoggerFactory loggerFactory, Exception exception)
+    protected override void LogException(ILogger logger, Exception exception)
     {
-        ILogger logger = loggerFactory.CreateLogger<MyExceptionHandlingMiddleware>();
         logger.LogCritical(exception, "Custom exception handling!");
     }
 
-    protected override HttpStatusCode DetermineResponseStatusCode(Exception exception)
+    protected override void WriteFailureToResponse(Exception exception, HttpStatusCode defaultFailureStatusCode, HttpContext context)
     {
         if (exception is ValidationException)
         {
-            return HttpStatusCode.BadRequest;
+            context.Response.StatusCode = (int) HttpStatusCode.BadRequest;
         }
-
-        if (exception is TimeoutException)
+        else if (exception is TimeoutException)
         {
-            return HttpStatusCode.ServerTimeout;
+            context.Response.StatusCode = (int) HttpStatusCode.ServerTimeout;
         }
-
-        return HttpStatusCode.InternalServerError;
+        else 
+        {
+            context.Response.StatusCode = (int) defaultFailureStatusCode;
+        }
     }
 }
 ```


docs/preview/features/logging.md
```diff
@@ -61,6 +61,57 @@ app.UseEndpoints(endpoints => endpoints.MapControllers());
...
+    {
+    }
+
+    protected override void LogException(ILoggerFactory loggerFactory, Exception exception)
+    {
+        ILogger logger = loggerFactory.CreateLogger<MyExceptionHandlingMiddleware>();
+        logger.LogCritical(exception, "Custom exception handling!");
+    }
+
+    protected override HttpStatusCode DetermineResponseStatusCode(Exception exception)
```
fgheysels, line 87: Instead of just allowing to customize the http status code that is returned, I think we should allow end-users (developers) to completely customize the error - response (not only the status code, but also the body)

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
```diff
@@ -62,31 +62,48 @@ public async Task Invoke(HttpContext context, ILoggerFactory loggerFactory)
...
+                context.Response.StatusCode = (int) statusCode;
+            }
+        }
+
+        /// <summary>
+        /// Logs the caught <paramref name="exception"/> before the response it sent back.
+        /// </summary>
+        /// <param name="loggerFactory">The factory instance to create <see cref="ILogger"/> instances, specifically for the <paramref name="exception"/>.</param>
+        /// <param name="exception">The caught exception during the application pipeline.</param>
+        protected virtual void LogException(ILoggerFactory loggerFactory, Exception exception)
```
fgheysels, line 79: Wouldn't it be easier for consumers to pass in an `ILogger` already, instead of the `ILoggerFactory`?

stijnmoreels: Ah, yes, it would be. 😄 Was wonder how much control users would want.

docs/preview/features/logging.md
stijnmoreels: Aha! Yeah, that could be useful. Also having access to the returned headers and all. 👍

stijnmoreels: Done in https://github.com/arcus-azure/arcus.webapi/pull/342/commits/08ede01488cef27e726f8cea3ed343379cce1ed3

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
stijnmoreels: Done in https://github.com/arcus-azure/arcus.webapi/pull/342/commits/5e59eb6082bc01739b9fc13a2195e8474614d505

fgheysels: I think in this case, users prefer simplicity :)

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
```diff
@@ -62,31 +62,47 @@ public async Task Invoke(HttpContext context, ILoggerFactory loggerFactory)
...
-            catch (Microsoft.AspNetCore.Server.Kestrel.Core.BadHttpRequestException ex) 
+            if (exception is Microsoft.AspNetCore.Server.Kestrel.Core.BadHttpRequestException badRequestException)
 #endif
             {
                 // Catching the `BadHttpRequestException` and using the `.StatusCode` property allows us to interact with the built-in ASP.NET components.
                 // When the Kestrel maximum request body restriction is exceeded, for example, this kind of exception is thrown.
 
-                LogException(loggerFactory, ex);
-                context.Response.StatusCode = ex.StatusCode;
+                context.Response.StatusCode = badRequestException.StatusCode;
```
fgheysels, line 102: Isn't this always overwritten by line 105 ?

fgheysels: Also, this (catching this BadRequestException) and retrieving the statuscode might be sometehing that is very useful, and ppl that override this default behaviour maybe want this as well.

I was thinking: maybe we can change this method, by extracting this logic outside of the virtual method and passing a 'default status code' as argument.  This default status code could then be by default 500, but if such an exception was thrown, we pass in that statuscode that is a property of the exception ?

Something like this (very roughly):

```csharp
catch( Exception ex )
{
    int defaultStatusCode = 500;

    if( exception is BadRequestException badReqException )
    {
        defaultStatusCode = badReqException.StatusCode;
    }

    WriteFailureToResponse(defaultStatusCode, ex, context);
}

protected virtual WriteFailureToResponse(int defaultHttpStatusCode, Exception ex, HttpContext context)
{
    context.Response.StatusCode = defaultHttpStatusCode;
}
````

Comment (stijnmoreels) [reply src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs:102]: Yeah that's true. Think this by-default catching of bad request is kinda useful and was a bit removed by fully placing it in the overridable method. Your suggestion might be a good way to have best of both worlds, yeah. Let's try something! 👍

Comment (stijnmoreels) [reply src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs]: Implemented something here: https://github.com/arcus-azure/arcus.webapi/pull/342/commits/3a0691f62bd4a6a9f799988e6ea1d755c5cde148

Comment (fgheysels) [reply src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs]: Great!

File: src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs [hunk 4]
```diff
@@ -63,30 +63,56 @@ public async Task Invoke(HttpContext context, ILoggerFactory loggerFactory)
...
 
-            logger.LogCritical(ex, ex.Message);
+        /// <summary>
+        /// Determine the HTTP status code based on the caught exception.
+        /// </summary>
+        /// <param name="exception">The caught exception during the application pipeline.</param>
+        /// <param name="context">The context instance for the current HTTP request.</param>
+        /// <param name="defaultFailureStatusCode">The default HTTP status code for the failure that was determined by the caught <paramref name="exception"/>.</param>
+        /// <returns>An HTTP status code that represents the <paramref name="exception"/>.</returns>
+        protected virtual void WriteFailureToResponse(Exception exception, HttpContext context, HttpStatusCode defaultFailureStatusCode)
```
fgheysels, line 113: Just not sure if the order of parameters is ok.
`exception` and `defaultFailureStatusCode` are both parameters which just provide you information whereas the `context` parameter is the one that you manipulate.   Maybe `context` should be the last parameter ?

docs/preview/features/logging.md
```diff
@@ -61,6 +61,57 @@ app.UseEndpoints(endpoints => endpoints.MapControllers());
...
+            return HttpStatusCode.BadRequest;
+        }
+
+        if (exception is TimeoutException)
+        {
+            return HttpStatusCode.ServerTimeout;
+        }
+
+        return HttpStatusCode.InternalServerError;
+    }
```
fgheysels, line 100: This sample will need to be adapted to reflect the new situation.

stijnmoreels: Ah, yes (stupid me 😅 ).

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
stijnmoreels: Done in https://github.com/arcus-azure/arcus.webapi/pull/342/commits/e6605a2f99528abc9e4c181d8a5e5b5b0ffcbdf3

docs/preview/features/logging.md
stijnmoreels: Done in https://github.com/arcus-azure/arcus.webapi/pull/342/commits/989708ae0bfd9dcdfb95e0304f0ec51e082144aa

src/Arcus.WebApi.Logging/ExceptionHandlingMiddleware.cs
fgheysels: I'm glad I'm not the only one who finds such small things important :D

docs/preview/features/logging.md
fgheysels: This looks good!

fgheysels (commented):
Looks good, thanks!
As soon as this is released, I might be able to introduce Arcus in another project :)

PR #345, opened 2022-06-24
feat: extract operation parent ID from request

### LINKED ISSUE #341: Extract operation parent ID from `Request-Id` instead of operation ID
Currently, we extract the root service ID from the `Request-Id` header which becomes the operation ID and the entire header value becomes the operation parent ID. This is wrong.

We should extract the latest ID (new dependency ID) from the `Request-Id` header and assign it as the operation parent ID. The operation ID should be generated (first the trace identifier is checked).

So an incoming `Request-Id`: `|abc.def` should result in:
- `OperationParentId` = `def`
- `OperationId` = `ghi` (generated)

stijnmoreels:
Extracts the operation parent ID from the `Request-Id` header instead of the operation ID. This was a fault in the correlation middleware component.
Now, we extract the operation parent ID from the request and generate a new operation ID for the telemetry that comes after, linking the requests with possible dependencies and other telemetry types.

Closes #341


docs/preview/features/correlation.md
```diff
@@ -22,7 +22,7 @@ This correlation is based on the `RequestId` and `X-Transaction-ID` HTTP request
 
 ## How This Works
 
-When an application is configured to use the default configuration of the correlation, each HTTP response will get an extra header called `RequestId` containing an unique identifier to distinguish between requests/responses.
+When an application is configured to use the default configuration of the correlation, each HTTP response will get an extra header called `RequestId` containing an unique identifier to distinguish between requests/responses. This ID will act as the parent ID for all telemetry that comes after and uses the [HTTP Correlation](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md) to extract the most recent ID.
```
fgheysels, line 25:
```suggestion
When an application is configured to use the default configuration of the correlation, each HTTP **response** will have an extra header called `RequestId` containing an unique identifier to distinguish between requests/responses. This ID will act as the parent ID for all telemetry that comes after and uses the [HTTP Correlation](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md) to extract the most recent ID.
```
Made it more clear that it is the response that has an extra header.
The sentencee that was added, is not quite clear imho though.   I think you mean that all requests to other dependencies inside the request, use this id as parent id ?

docs/preview/features/correlation.md
```diff
@@ -22,7 +22,7 @@ This correlation is based on the `RequestId` and `X-Transaction-ID` HTTP request
...
 
 The `X-Transaction-ID` can be overridden by the request, meaning: if the HTTP request already contains a `X-Transaction-ID` header, the same header+value will be used in the HTTP response.
```
fgheysels, line 27: Shouldn't we  first explain what this `x-transaction-id` header is about ?

fgheysels: Something like:

```suggestion
The `X-Transaction-ID` can be overridden by the request, meaning: if the HTTP request already contains a `X-Transaction-ID` header, the same header+value will be used in the HTTP response.
This header is used to group all logs that belong to the same functional operation together.  In other words, using this ID, you should be able to find all logs that were generated for this logical operation.
```

docs/preview/features/correlation.md
stijnmoreels: Yes, although I had problems reading your comment 😄 .

docs/preview/features/correlation.md
stijnmoreels: Yes, I like it. 👍

docs/preview/features/correlation.md
```diff
@@ -22,9 +22,10 @@ This correlation is based on the `RequestId` and `X-Transaction-ID` HTTP request
 
 ## How This Works
 
-When an application is configured to use the default configuration of the correlation, each HTTP response will get an extra header called `RequestId` containing an unique identifier to distinguish between requests/responses.
+When an application is configured to use the default configuration of the correlation, each HTTP **response** will have an extra header called `RequestId` containing an unique identifier to distinguish between requests/responses. This ID will act as the parent ID for all telemetry that comes after and uses the [HTTP Correlation](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md) to extract the most recent ID.
```
fgheysels, line 25:
```suggestion
When an application is configured to use the default configuration of the correlation, each HTTP **response** will have an extra header called `RequestId` containing an unique identifier to distinguish between requests/responses. This ID will be used by the [HTTP Correlation](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md) as the parent ID for all the requests that are triggered by this operation.
```

stijnmoreels: 'requests' seems misleading, no? In what contexts are there more than 1 requests linked to a single dependency?

stijnmoreels: In any case, I would want us to add a user guide for observability that explains fully the service-to-service correlation and how Web API and Messaging can be used in combination for this.

fgheysels: Indeed, I was doubting how to write it.
I actually meant 'requests initiated by calls to dependencies from this operation'.   So, your operation has a dependency on service X, and initiates a 'request' to Service X.   Service X will receive that request, and will log it as a request.

stijnmoreels: Maybe in this context, we shouldn't use 'request' and 'dependency' in a not telemetry type way.
I mean, not say: '... has a dependency on...', '...sends request too...', ... bc it becomes really confusing. Since it's multiple services, the viewpoint also alters the name. In the sending service it is a dependency, in the receiving service it is a request. Maybe a visual can help here.

stijnmoreels: Rewrote the entire section and added a visual here: https://github.com/arcus-azure/arcus.webapi/pull/345/commits/92657efdc2aa14b157ba08761c31945565a5e424

fgheysels: Great! Thanks.

fgheysels (approved):

Issue #329, opened 2022-06-01
Change target-type of Security related extension methods


fgheysels:
We create some extension methods to easily add our custom authentication filters to the ASP.NET pipeline.

These extension methods (for instance `AddSharedAccessKeyAuthenticationOnHeader`) are extension methods on the `FilterCollection` type.  This means we have to call them like this:

```csharp
builder.Services.AddControllers( options => 
    {
        options.Filters.AddSharedAccessKeyAuthenticationOnHeader( ... );
    });
```

Suggestion would be to change these methods so that they extend the `MvcOptions` type.  By doing so, they can be used like this:

```csharp
builder.Services>.AddControllers( options =>
    {
        options.AddSharedAccessKeyAuthenticationOnHeaderFilter( ... );
    });
```

stijnmoreels:
`...OnHeaderFilter` or `...FilterOnHeader` 😅 ?

stijnmoreels:
All changed! 🥳

PR #353, opened 2022-07-06
feat: change jwt target type

stijnmoreels:
Change target type of JWT authorization filter to the MVC options instead.

Relates to #329


fgheysels (approved):
I think this is a nice improvement for developer experience :)

PR #361, opened 2022-07-20
chore: make http correlation az func available

stijnmoreels:
Use the empty HTTP correlation extension in Azure Functions.


stijnmoreels:
Think this is allowed, even though, we don't work with a major version.

fgheysels (approved):

PR #366, opened 2022-07-26
fix: correct default transaction id header name

stijnmoreels:
Correct default transaction ID for both sending and receiving options.


stijnmoreels:
Fixed in #367

PR #371, opened 2022-08-04
(❗) feat: add telemetry context to http client correlation options

### LINKED ISSUE #370: Add telemetry context option to HTTP client correlation tracking options
Currently, when sending a HTTP request with the HTTP client correlation tracking, we aren't able to send any additional telemetry data, while in certain cases this is required (see Azure Functions (in-process)) or very useful for additional information.

Add an option that let's you a telemetry context to the options. We shouldn't expose directly the `IDictionary<>` as this is a bad practice, we could add a method on the options model to add such a context with a method, like:

```csharp
options.AddTelemetryContext(...);
```

**Additional context**
Discovered by question of @pim-simons

stijnmoreels:
Enhances HTTP client correlation options to pass telemetry context so that the following tracked HTTP dependency has additional custom dimensions.

Closes #370

### DIFF
--- a/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
+++ b/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
@@ -61,7 +61,7 @@ namespace Arcus.WebApi.Logging.Core.Correlation
         /// <summary>
         /// Gets the telemetry context used during HTTP dependency tracking.
         /// </summary>
-        internal Dictionary<string, object> TelemetryContext { get; private set; } = new Dictionary<string, object>();
+        internal Dictionary<string, object> TelemetryContext { get; } = new Dictionary<string, object>();
 
         /// <summary>
         /// Adds a telemetry context while tracking the HTTP dependency.
--- a/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
+++ b/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
@@ -73,7 +73,7 @@ namespace Arcus.WebApi.Logging.Core.Correlation
             Guard.NotNull(telemetryContext, nameof(telemetryContext), "Requires a telemetry context dictionary to add to the HTTP dependency tracking");
             foreach (KeyValuePair<string, object> item in telemetryContext)
             {
-                TelemetryContext.Add(item.Key, item.Value);
+                TelemetryContext[item.Key] = item.Value;
             }
         }
     }
--- a/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
+++ b/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Net.Http;
 using GuardNet;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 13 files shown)


src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
```diff
@@ -56,5 +57,21 @@ public string TransactionIdHeaderName
...
+        /// Gets the telemetry context used during HTTP dependency tracking.
+        /// </summary>
+        internal Dictionary<string, object> TelemetryContext { get; private set; } = new Dictionary<string, object>();
+
+        /// <summary>
+        /// Adds a telemetry context while tracking the HTTP dependency.
+        /// </summary>
+        /// <param name="telemetryContext">The dictionary with contextual information about the dependency telemetry.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="telemetryContext"/> is <c>null</c>.</exception>
+        public void AddTelemetryContext(Dictionary<string, object> telemetryContext)
```
fgheysels, line 71: Would it be beneficial to have an overload that just takes a key/value pair, instead of a dictionary ?

Question that arises: what happens when using that overload in combination with the method that takes a complete dictionary.
IMHO, this method should not set / overwrite the `TelemetryContext` property.  Instead, it should take the key/value pairs from the `telemetryContext` parameter, and set them in the `TelemetryContext` property.  (If an entry already exists: overwrite it ?)

stijnmoreels: Usually, when this is used, there is already a telemetry context available that is being used for other tracking as well, therefore it would be cumbersome to only have key/value pair instead of a dictionary. We could, however, when the need arises go that path.

Adding instead of overwritting, that could indeed be an improvement. 👍

stijnmoreels:
For some reason, the CodeQL code analysis fails on missing .NET SDK's 😢 
Will have to look another time with a clearer head. 😄

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
fgheysels: I meant to have an overload that only accepts key/value pairs next to the method that is introduced here.  But I can see that it might not be necessary

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelationClientOptions.cs
```diff
@@ -56,5 +57,24 @@ public string TransactionIdHeaderName
...
+        {
+            Guard.NotNull(telemetryContext, nameof(telemetryContext), "Requires a telemetry context dictionary to add to the HTTP dependency tracking");
+            foreach (KeyValuePair<string, object> item in telemetryContext)
+            {
+                TelemetryContext.Add(item.Key, item.Value);
```
fgheysels, line 76: What if there already exists an entry for `item.Key` ?  I think we should check on that, and decide if we overwrite or not (I'd say overwrite)

stijnmoreels: Ah, then we could just use `dictionary[key] = value` for that, as that is the default behavior. The `Add` will throw.

stijnmoreels: Done in https://github.com/arcus-azure/arcus.webapi/pull/371/commits/0780250730751c017a507e32622d157f263c0d8b

fgheysels (approved):

PR #372, opened 2022-08-05
(❗) chore: update missing request id warning to trace

### LINKED ISSUE #369: Correlation tracing in application insights
**Describe the bug**
My customer is running HTTP Triggered Azure Functions and logging telemetry using Arcus.Observability to Application Insights. 
We are now getting a lot of trace information that we did not get before in Application Insights containing this information:
image

We don't need/want this information in Application Insights but I cannot find a way to get rid of these traces. 
This seems to have originated when updating to Arcus.WebApi.Logging.AzureFunctions 1.6.0. The code of the application did not change, only the Arcus packages were updated.

Did something change so that these trace warnings are now logged? And is there a way to prevent these trace messages being logged?

**To Reproduce**
Run an HTTP Triggered Azure Function and log information to Application Insights.

**Expected behavior**
These messages are not expected to be logged to Application Insights:
image

stijnmoreels:
Use `Trace` io `Warning` to write diagnostic message about missing `Request-Id` as this is sometimes correct (like in the initial request).

Closes #369

### DIFF
--- a/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelation.cs
+++ b/src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelation.cs
@@ -361,7 +361,7 @@ namespace Arcus.WebApi.Logging.Correlation
                 {
                     if (string.IsNullOrWhiteSpace(requestId))
                     {
-                        _logger.LogWarning("No response header was added given no operation parent ID was found");
+                        _logger.LogTrace("No response header was added given no operation parent ID was found");
                     }
                     else
                     {


src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelation.cs
```diff
@@ -361,7 +361,7 @@ private void AddCorrelationResponseHeaders(HttpContext httpContext, string reque
                 {
                     if (string.IsNullOrWhiteSpace(requestId))
                     {
-                        _logger.LogWarning("No response header was added given no operation parent ID was found");
+                        _logger.LogTrace("No response header was added given no operation parent ID was found");
```
fgheysels, line 364: Am I missing something ?  The related issue talks about another log-entry that should have a 'trace' level, but it is not included in the PR ?

fgheysels (approved):

src/Arcus.WebApi.Logging.Core/Correlation/HttpCorrelation.cs
stijnmoreels: The issue was still using v1.6, while this is already fixed in the preview version of v1.6.1-preview-1; which is also in the issue discussion:

> Upgraded to version v1.6.1-preview-1 and did a test, the Operation correlation ID's cannot be added to response, but the Operation.IncludeInResponse setting was still set trace messages has indeed disappeared.

Issue #358, opened 2022-07-08
Extra overload for SharedAccessKeyAttribute


gverstraete:
**Is your feature request related to a problem? Please describe.**
The current code here: https://github.com/arcus-azure/arcus.webapi/blob/master/src/Arcus.WebApi.Security/Authentication/SharedAccessKey/SharedAccessKeyAuthenticationAttribute.cs

supports Azure Keyvault integration which is great! but I'd like an extension to this.

**Describe the solution you'd like**
We are working on a key rotation system, which means that when a new secret version is deployed, the new one & the old one should be still active & be considered as "valid"..

To make it easy, I would make the amount of versions to validate configurable.

Idealy this works together with the Event Grid event when a new version is deployed.

In this way we can have rolling updates on our secrets & api-keys.

Let me know what you think.

stijnmoreels:
Hi Gillian, thx for mentioning.
Have you looked at [this documentation page](https://background-jobs.arcus-azure.net/Features/Security/auto-invalidate-secrets) where we talk about automatic notification of the cache. It also works with the Arcus store, so it can be used for shared access key authentication. The authentication mechanism will call the cached variant from the secret store automatically, and you could notify with Key Vault events when a certain secret should be invalidated by the cache. A background job will then notify the secret store, and eventually, the shared access key authentication will ask for a different secret.

gverstraete:
Hi @stijnmoreels !

Yes, I did look at that documentation :). The main point that I want to make here is that we should be able to have two valid versions of the secret. The last one & the last -1.

Both should be validated with that attribute, and that is something that will need to be added...

It's like in storage accounts, there you have primary & secondary keys. you rotate by updating one of the two, and migrating all your applications to the new one. I want to do something similar here with keyvault secrets.

stijnmoreels:
Aha, ok, hmm. Think this is more something we should add in our Arcus secret store, and not in Web API, as the Azure Key Vault integration here goes via the secret store.

gverstraete:
Yes, but I'd prefer to only have one secret, because secrets have versioning in keyvault.

stijnmoreels:
Done in #380

stijnmoreels:
@gverstraete , we will close this bc it is implemented and provided with a preview version. Feel free to re-open it in case of any problems.

PR #380, opened 2022-09-07
feat: add support for versioned secrets in shared access key auth

stijnmoreels:
Add support for versioned secrets in the shared access key authentication.

### DIFF
--- a/docs/preview/03-Features/security/auth/shared-access-key.md
+++ b/docs/preview/03-Features/security/auth/shared-access-key.md
@@ -67,7 +67,7 @@ builder.Services.AddControllers(options =>
 For this setup to work, an Arcus secret store is required as the provided secret name (in this case `"shared-access-key-name"`) will be looked up.
 See [our official documentation](https://security.arcus-azure.net/features/secret-store/) for more information about setting this up.
 
-> 💡 This authentication mechanism supports multiple secret versions so that many different versions of a single configured secret are checked during authentication. See [our official documentation]() for more information on how to configured your secret as 'versioned' in the secret store.
+> 💡 This authentication mechanism supports multiple secret versions so that many different versions of a single configured secret are checked during authentication. See [our official documentation](https://security.arcus-azure.net/Features/secret-store/versioned-secret-provider) for more information on how to configured your secret as 'versioned' in the secret store.
 
 ## Enforce shared access key authentication per controller or operation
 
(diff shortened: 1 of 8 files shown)


stijnmoreels:
After this, we could provide a preview package so that Gillian already can take a look.

docs/preview/03-Features/security/auth/shared-access-key.md
```diff
@@ -79,6 +67,8 @@ builder.Services.AddControllers(options =>
 For this setup to work, an Arcus secret store is required as the provided secret name (in this case `"shared-access-key-name"`) will be looked up.
 See [our official documentation](https://security.arcus-azure.net/features/secret-store/) for more information about setting this up.
 
+> 💡 This authentication mechanism supports multiple secret versions so that many different versions of a single configured secret are checked during authentication. See [our official documentation]() for more information on how to configured your secret as 'versioned' in the secret store.
```
stijnmoreels, line 70: Build failed bc we can't fill this out yet, as the security docs are not yet released: https://github.com/arcus-azure/arcus.security/pull/363

fgheysels:
To be honest, I'm currently really wondering what kind of problem this 'versioned secret' thing solves.
When you update a certain key for instance, the consumer of the resource for which you need that key, will need to use the most recent secret.   I don't see why you would -in a normal situation- need to fall back to an older version of a secret ?

stijnmoreels:
> To be honest, I'm currently really wondering what kind of problem this 'versioned secret' thing solves. When you update a certain key for instance, the consumer of the resource for which you need that key, will need to use the most recent secret. I don't see why you would -in a normal situation- need to fall back to an older version of a secret ?

This mostly came from a feature request from @gverstraete, https://github.com/arcus-azure/arcus.webapi/issues/358

fgheysels (approved):
I approve this, although I don't see much use for this feature to be honest.
When a new version of a secret is added in KeyVault, you typically want to use that new version of a secret, and not the previous one.
When a new version is added, it means for instance the key that must be used to access a certain resource has been updated.  So I don't see no reason to be able to fall back on the previous version of the scret :?

stijnmoreels:
> I approve this, although I don't see much use for this feature to be honest. When a new version of a secret is added in KeyVault, you typically want to use that new version of a secret, and not the previous one. When a new version is added, it means for instance the key that must be used to access a certain resource has been updated. So I don't see no reason to be able to fall back on the previous version of the scret :?

Think this is the main request that is being supported here: "We are working on a key rotation system, which means that when a new secret version is deployed, the new one & the old one should be still active & be considered as "valid".."

PR #381, opened 2022-09-08
chore: simplify certificate authentication

stijnmoreels:
Simplify the certificate validation used in the certificate authentication. This includes correct logging and removal of unnecessary internal domain types.


src/Arcus.WebApi.Security/Authentication/Certificates/CertificateAuthenticationConfig.cs
```diff
@@ -14,7 +14,7 @@ namespace Arcus.WebApi.Security.Authentication.Certificates
     /// </summary>
     public class CertificateAuthenticationConfig
     {
-        private readonly IDictionary<X509ValidationRequirement, (IX509ValidationLocation location, ConfiguredKey configuredKey)> _locationAndKeyByRequirement;
+        private readonly IDictionary<X509ValidationRequirement, (IX509ValidationLocation location, string configuredKey)> _locationAndKeyByRequirement;
```
fgheysels, line 17: I'm just wondering ....
In the past you were totally convinced of using specific 'domain types' instead of using primitives, but in this case you're moving away from them.  Why ? What is the reasoning ?

fgheysels: I might see the answer already, as the `ConfiguredKey` type does not really provide any added value :)

fgheysels (approved):

src/Arcus.WebApi.Security/Authentication/Certificates/CertificateAuthenticationConfig.cs
stijnmoreels: Yes, 😄 , I'm still totally convinced that specific domain types are the way to go if you want any kind of secure and maintainable application. In this case, however, we're talking about relationship between two internal methods, so the added benefits are rather low as we don't even reuse the domain validation, here.

stijnmoreels: I've written even numerous blog posts on this matter, on how to set those up and how to use them practically. 'Secure by design' is the idea here. 😄

PR #387, opened 2022-09-21
feat: simplify cert auth validation registration

### LINKED ISSUE #350: Make the `CertificateAuthenticationValidator` configuration part of the certificate authentication registration
Currently, when registering the certificate authentication, the consumer is also required to register separately an `CertificateAuthenticationValidtor` that is responsible to  validate certain configured parts of the received client certificate.
This additional registration is not something that is clear when registering the certificate authentication filter.

Make the additional registration of the `CertificateAuthenticationValidator` explicit by making it part of the certificate authentication registration. This change could also make sure that we may not even have to register the validator in the application services as it can be passed along with the certificate authentication filter directly.

stijnmoreels:
Simplify the registration of the certificate authentication validator by either passing the configuration directly during the global certificate authentication filter or by using the new extension.

Closes #350

### DIFF
--- a/docs/preview/03-Features/security/auth/certificate.md
+++ b/docs/preview/03-Features/security/auth/certificate.md
@@ -13,17 +13,6 @@ This authentication process consists of following parts:
 2. Determine which properties of the received client certificate are used for authentication
 3. The property value(s) of the client certificate matches the value(s) determined via configured secret provider, configuration or custom implementation
 
-- [Authentication with certificate](#authentication-with-certificate)
-  - [Installation](#installation)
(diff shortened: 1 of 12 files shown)


fgheysels (approved):
Looks great. I've never had to use this in a project, but it's good to know that we have now this fine grained kind of control.

stijnmoreels:
> Looks great. I've never had to use this in a project, but it's good to know that we have now this fine grained kind of control.

Practically, there's no change in functionality here, but it is indeed made a lot simpler to add this kind of authentication.

Issue #309, opened 2022-03-16
AppInsights is transitioning to W3C TraceContext


fgheysels:
As can be read [here](https://docs.microsoft.com/en-us/azure/azure-monitor/app/correlation#correlation-headers-using-w3c-tracecontext), Application Insights is transitioning to TraceContext standard.

This means that some headers like `Request-Id' will be deprecated as far as I understand from the above article.

I believe that we will have some work to do in order to support this change, since I've seen that our `CorrelationMiddleware` in Arcus.WebApi is using the `CorrelationInfoUpstreamServiceOptions` to correlate, where the default header-name that is used to correlate is `Request-Id`.

Although the article states that backward compatiblity will be maintained, I feel we should defnitily investigate what these changes mean to the functionality provided by Arcus, and we must consider moving to this W3C std as well.

To be further investigated.

stijnmoreels:
### Current situation
When a request is received, following headers are inspected:
* `X-Transaction-Id`: becomes the `TransactionId` in the correlation
* `Request-Id`: contains both the operation and operation parent ID:
    * When matches the correct format, returns the root ID as the `OperationId`
    * Otherwise use the `X-Operation-Id` header for the `OperationId`
    * The entire value is used as the `OperationParentId` in the correlation

### New situation
When a request is received, following headers should be inspected:
* `traceparent`: contains both the transaction ID and the operation parent ID
    * `trace-id` becomes `OperationId`
    * `parent-id` becomes `OperationParentId`
* `tracestate`: contains a set of encoded sets of the `traceparent` and seems like a good one for the `TransactionId` 

That's how I saw it, but maybe we should look into Microsoft's code and take a peek how they do it.

PR #395, opened 2022-10-24
feat: add http func w3c correlation support

stijnmoreels:
Add W3C HTTP correlation support to the Azure Functions HTTP trigger environment.

Relates to #309


stijnmoreels:
Fixing weird test results on a later time with a clear head.

src/Arcus.WebApi.Tests.Runtimes.AzureFunction/Startup.cs
```diff
@@ -16,7 +15,7 @@ public class Startup : FunctionsStartup
         /// <param name="builder">The instance to build the registered services inside the functions app.</param>
         public override void Configure(IFunctionsHostBuilder builder)
         {
-            builder.AddHttpCorrelation(configureOptions: (Action<HttpCorrelationInfoOptions>) null);
+            builder.AddHttpCorrelation(options => options.Format = HttpCorrelationFormat.Hierarchical);
```
stijnmoreels, line 18: Couldn't do any better. Had some problems with Docker <> W3C <> Azure Functions, so changed this to hierarchical for testing. The service is tested separately.

fgheysels: Weird, what kind of problem ?

stijnmoreels: I got it to work manually, but via the YAML build, I keep having a `TelemetryClient` not found, or a `.dll` not found exception. Tried several things for more than a week, so updated this eventually to hierarchical. It is tested separately, and is also in-process (which will be deprecated).

src/Arcus.WebApi.Logging.AzureFunctions/Correlation/AzureFunctionsInProcessHttpCorrelation.cs
```diff
@@ -0,0 +1,114 @@
...
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Arcus.WebApi.Logging.AzureFunctions.Correlation
+{
+    /// <summary>
+    /// Represents an <see cref="HttpCorrelationTemplate{THttpRequest,THttpResponse}"/> implementation
+    /// that extracts and sets HTTP correlation throughout Azure Functions (in-process) HTTP trigger applications.
```
fgheysels, line 17: Why is there a difference in in-process & out-of-process correlation ?

src/Arcus.WebApi.Tests.Runtimes.AzureFunction/Startup.cs
fgheysels: Weird, are we sure we're not going to run into this when using it in a real project ?

src/Arcus.WebApi.Tests.Core/TraceParent.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            string transactionId = BogusGenerator.Random.Hexadecimal(32, prefix: null);
+            string operationParentId = BogusGenerator.Random.Hexadecimal(16, prefix: null);
+
+            return new TraceParent(transactionId, operationParentId);
+        }
+
+        public override string ToString()
+        {
+            return $"00-{TransactionId}-{OperationParentId}-00";
```
fgheysels, line 33: Just wondering what these 00 prefix & suffix represent.

src/Arcus.WebApi.Logging.AzureFunctions/Correlation/AzureFunctionsInProcessHttpCorrelation.cs
stijnmoreels: Because of the type and hosting functionality it uses.
In-process uses the same `HttpRequest`/`HttpResponse` types, but does not have middleware like a regular ASP.NET Core application.
Isolated uses `HttpRequestData`/`HttpResponseData` types, but has middleware like a regular ASP.NET Core application.

src/Arcus.WebApi.Tests.Core/TraceParent.cs
stijnmoreels: Version and structure.
https://www.w3.org/TR/trace-context/

src/Arcus.WebApi.Tests.Runtimes.AzureFunction/Startup.cs
stijnmoreels: I got it to work locally without Docker, as my Docker knowledge is too limited very this. I think it has to do with the restore operation of packages and that it uses different versions when building via Docker. But some help would be helpful. Feel free to check this yourself. 👍

fgheysels: > and that it uses different versions when building via Docker

That would be very strange.
I suggest that we complete this PR and look at it together; that might be most productive.  Maybe when we're both at the office ?

fgheysels (approved):

src/Arcus.WebApi.Tests.Runtimes.AzureFunction/Startup.cs
stijnmoreels: My days at the office are still very limited and I can't say for sure when the next time will be. Definitely a couple weeks from now. It is the remote build that is failing, so the only errors are the logs from the build machine.

stijnmoreels: Ok, I think I found something. It has something to do with an internal conflict of app insights packages.
in-process Azure Functions may need to use this package: https://learn.microsoft.com/en-us/azure/azure-functions/functions-dotnet-class-library?tabs=v2%2Ccmd#log-custom-telemetry-in-c-functions

fgheysels: Great find.  Eager to know if that fixes the issue or not.

stijnmoreels: It does, and the change has already be pushed here.

PR #406, opened 2022-11-23
fix: relaxed allow header json formatting

### LINKED ISSUE #405: JSON formatting middleware should check if Accept header contains JSON || ANY instead of exact equal
**Describe the bug**
When an Accept header contain any additional values (like `q=4`) together with the media type, then the JSON middleware component (both Azure Functions) only checks for an exact match while `application/json, q=4` should also still match.

**To Reproduce**
Run project and try to navigate towards public OpenAPI endpoint in an Azure Functions project.

**Expected behavior**
Public OpenAPI endpoint is shown.

**Additional context**
 - v1.7.0-preview-3

stijnmoreels:
We need to make sure that we also accept headers with a factor weighting (`q=`).

Closes #405

### DIFF
--- a/src/Arcus.WebApi.Hosting.AzureFunctions/Formatting/AzureFunctionsJsonFormattingMiddleware.cs
+++ b/src/Arcus.WebApi.Hosting.AzureFunctions/Formatting/AzureFunctionsJsonFormattingMiddleware.cs
@@ -80,7 +80,7 @@ namespace Arcus.WebApi.Hosting.AzureFunctions.Formatting
         {
             if (request.Headers.TryGetValues("Accept", out IEnumerable<string> headerValues))
             {
-                if (headerValues.Any(value => value == "application/json" || value == "*/*"))
+                if (headerValues.Any(value => value?.Contains("application/json") is true || value?.Contains("*/*") is true))
                 {
                     return true;
                 }
--- a/src/Arcus.WebApi.Tests.Unit/Hosting/Formatting/AzureFunctionsJsonFormattingMiddlewareTests.cs
+++ b/src/Arcus.WebApi.Tests.Unit/Hosting/Formatting/AzureFunctionsJsonFormattingMiddlewareTests.cs
@@ -2,8 +2,8 @@
 using System.Text;
 using System.Threading.Tasks;
 using Arcus.WebApi.Hosting.AzureFunctions.Formatting;
-using Arcus.WebApi.Logging.AzureFunctions;
 using Arcus.WebApi.Tests.Unit.Logging.Fixture.AzureFunctions;
+using Bogus;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Azure.Functions.Worker.Http;
 using Xunit;
@@ -12,6 +12,8 @@ namespace Arcus.WebApi.Tests.Unit.Hosting.Formatting
 {
     public class AzureFunctionsJsonFormattingMiddlewareTests
     {
+        private static readonly Faker BogusGenerator = new Faker();
+
         [Fact]
         public async Task Request_WithoutJsonFormattingHeaders_ReturnsFailure()
         {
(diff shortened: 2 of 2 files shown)


src/Arcus.WebApi.Tests.Unit/Hosting/Formatting/AzureFunctionsJsonFormattingMiddlewareTests.cs
```diff
@@ -131,6 +131,42 @@ public async Task Request_WithAllAllowAndJsonContentTypeHeaders_ReturnsOk()
...
         }
 
+        [Fact]
+        public async Task Request_WithJsonAllowHeaderWithExtension_ReturnsOk()
+        {
+            // Arrange
+            var middleware = new AzureFunctionsJsonFormattingMiddleware();
+            var context = TestFunctionContext.Create(req =>
+            {
+                req.Headers.TryAddWithoutValidation("allow", "application/json, q=4");
```
fgheysels, line 141: q value can only range between 0 or 1 :)  Ok, this is a test so it doesn't matter in fact.
See also here on how it is used:
https://restfulapi.net/q-parameter-in-http-accept-header/

src/Arcus.WebApi.Hosting.AzureFunctions/Formatting/AzureFunctionsJsonFormattingMiddleware.cs
```diff
@@ -80,7 +80,7 @@ private static bool AcceptsJson(HttpRequestData request, ILogger logger)
         {
             if (request.Headers.TryGetValues("Accept", out IEnumerable<string> headerValues))
             {
-                if (headerValues.Any(value => value == "application/json" || value == "*/*"))
+                if (headerValues.Any(value => value?.Contains("application/json") is true || value?.Contains("*/*") is true))
```
fgheysels, line 83: When I think about this, why do we by default restrict that clients can only understand json ? 
Because this is really what the accept header is used for.  The client indicates what kind of mime-type the client can accept.  This means that the server can use this information to decide in what format the request should be returned.
So, isn't it a bit weird that we enforce that this can only be json by default ?  I think we (Arcus) should not intervene here ?

In fact, it is weird that we accept 'json' or  everything ('*/*').

fgheysels (approved):

src/Arcus.WebApi.Tests.Unit/Hosting/Formatting/AzureFunctionsJsonFormattingMiddlewareTests.cs
stijnmoreels: Woops, 😅

src/Arcus.WebApi.Hosting.AzureFunctions/Formatting/AzureFunctionsJsonFormattingMiddleware.cs
stijnmoreels: This is actually an alternative to the input/output formaters in a Web API. There we also have restricted the input and output to only JSON. In Azure Functions we do not have that possibility, so we created it ourselves.
So, this is actually already there in Web API, and was also already in the in-process Azure Functions, we only now were able to do this in isolated Azure Functions.

src/Arcus.WebApi.Tests.Unit/Hosting/Formatting/AzureFunctionsJsonFormattingMiddlewareTests.cs
stijnmoreels: I'll change that.

stijnmoreels: Done in https://github.com/arcus-azure/arcus.webapi/pull/406/commits/8cd193eb9bf81703f1aa2f5a53cc436e8869ef94

src/Arcus.WebApi.Hosting.AzureFunctions/Formatting/AzureFunctionsJsonFormattingMiddleware.cs
fgheysels: I just ask myself why Arcus needs to decide that.  But that's another discussion :)

fgheysels (approved):

PR #410, opened 2022-11-30
fix: LogRequest writes the RouteTemplate as operation-name

### LINKED ISSUE #407: Arcus request middleware should log requests without specific route-parameters
I'm testing Arcus.WebApi.Logging 1.7.0-preview-3, more specifically the request logging middleware and I'm writing to application insights.

When I perform some requests to my API, those requests appear in the 'Performance' overview in Application Insights like this:

image

This is not good, as this means that the 'Performance' overview will contain an entry for each request where the route-parameters are different.

When you would use the default Microsoft middleware for requestlogging, the requests are logged like this:

image

This is better, as all requests to the same operation are grouped together, regardless the route-parameters.

fgheysels:
When logging API requests, we should not log the exact route containing the actual route-parameters.  Instead, we should log the route-template as operation-name so that requests to the same endpoint are grouped together in Application Insights, regardless of their exact route-parameters.

Closes #407

### DIFF
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -220,7 +220,7 @@ namespace Arcus.WebApi.Logging
         {
             string operationName = null;
 
-            if (EndpointHttpContextExtensions.GetEndpoint(httpContext) is RouteEndpoint re)
+            if (httpContext.Features.Get<Microsoft.AspNetCore.Http.Features.IEndpointFeature>()?.Endpoint is RouteEndpoint re)
             {
                 operationName = re.RoutePattern.RawText;
             }
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -1,21 +1,15 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Net;
 using System.Threading.Tasks;
 using Arcus.Observability.Telemetry.Core;
 using Arcus.WebApi.Logging.Core.RequestTracking;
 using GuardNet;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Primitives;
-using Serilog.Context;
 
 namespace Arcus.WebApi.Logging
 {
@@ -218,14 +212,14 @@ namespace Arcus.WebApi.Logging
 
         private static string DetermineRequestOperationName(HttpContext httpContext)
         {
-            string operationName = null;
-
             if (httpContext.Features.Get<IEndpointFeature>()?.Endpoint is RouteEndpoint routingEndpoint)
             {
-                operationName = routingEndpoint.RoutePattern.RawText;
+                string operationName = routingEndpoint.RoutePattern.RawText;
+
+                return $"{httpContext.Request.Method} {operationName}";
             }
 
-            return $"{httpContext.Request.Method} {operationName}";
+            return null;
         }
 
         private async Task<string> GetPotentialRequestBodyAsync(HttpContext httpContext, bool includeRequestBody)
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -12,6 +12,7 @@ using GuardNet;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using Serilog.Context;
@@ -210,7 +211,26 @@ namespace Arcus.WebApi.Logging
             Dictionary<string, object> telemetryContext = CreateTelemetryContext(requestBody, responseBody, httpContext.Request.Headers, _logger);
             telemetryContext.Add("Body", "Request body is now available in 'RequestBody' dimension");
 
-            _logger.LogRequest(httpContext.Request, httpContext.Response, duration, telemetryContext);
+            var operationName = BuildOperationName(httpContext);
+
+            _logger.LogRequest(httpContext.Request, httpContext.Response, operationName, duration, telemetryContext);
+        }
+
+        private static string BuildOperationName(HttpContext httpContext)
+        {
+            string operationName = null;
+
+            if (EndpointHttpContextExtensions.GetEndpoint(httpContext) is RouteEndpoint re)
+            {
+                operationName = re.RoutePattern.RawText;
+            }
+
+            if (String.IsNullOrWhiteSpace(operationName))
+            {
+                operationName = httpContext.Request.Path;
+            }
+
+            return $"{httpContext.Request.Method} {operationName}";
         }
 
         private async Task<string> GetPotentialRequestBodyAsync(HttpContext httpContext, bool includeRequestBody)
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -211,12 +211,12 @@ namespace Arcus.WebApi.Logging
             Dictionary<string, object> telemetryContext = CreateTelemetryContext(requestBody, responseBody, httpContext.Request.Headers, _logger);
             telemetryContext.Add("Body", "Request body is now available in 'RequestBody' dimension");
 
-            var operationName = BuildOperationName(httpContext);
+            var operationName = DetermineRequestOperationName(httpContext);
 
             _logger.LogRequest(httpContext.Request, httpContext.Response, operationName, duration, telemetryContext);
         }
 
-        private static string BuildOperationName(HttpContext httpContext)
+        private static string DetermineRequestOperationName(HttpContext httpContext)
         {
             string operationName = null;
 
--- a/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
+++ b/src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
@@ -220,14 +220,9 @@ namespace Arcus.WebApi.Logging
         {
             string operationName = null;
 
-            if (httpContext.Features.Get<Microsoft.AspNetCore.Http.Features.IEndpointFeature>()?.Endpoint is RouteEndpoint re)
+            if (httpContext.Features.Get<IEndpointFeature>()?.Endpoint is RouteEndpoint routingEndpoint)
             {
-                operationName = re.RoutePattern.RawText;
-            }
-
-            if (String.IsNullOrWhiteSpace(operationName))
-            {
-                operationName = httpContext.Request.Path;
+                operationName = routingEndpoint.RoutePattern.RawText;
             }
 
             return $"{httpContext.Request.Method} {operationName}";
--- a/src/Arcus.WebApi.Tests.Integration/Logging/Controllers/EchoController.cs
+++ b/src/Arcus.WebApi.Tests.Integration/Logging/Controllers/EchoController.cs
@@ -7,6 +7,11 @@ namespace Arcus.WebApi.Tests.Integration.Logging.Controllers
     {
         public const string GetPostRoute = "echo";
 
+        public static string GetPostRouteWithRouteParameters(int deviceId)
+        {
+            return $"devices/{deviceId}/echo";
+        }
+
         [HttpGet]
         [Route(GetPostRoute)]
         public IActionResult Get()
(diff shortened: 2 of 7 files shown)


src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -210,7 +211,26 @@ private void LogRequest(string requestBody, string responseBody, HttpContext htt
...
+            {
+                operationName = re.RoutePattern.RawText;
+            }
+
+            if (String.IsNullOrWhiteSpace(operationName))
+            {
+                operationName = httpContext.Request.Path;
+            }
+
+            return $"{httpContext.Request.Method} {operationName}";
```
stijnmoreels, line 233: We can just leave it `null` then the `{method} {path}` will automatically be filled-out as `operationName`.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -210,7 +211,26 @@ private void LogRequest(string requestBody, string responseBody, HttpContext htt
...
+        }
+
+        private static string BuildOperationName(HttpContext httpContext)
+        {
+            string operationName = null;
+
+            if (httpContext.Features.Get<Microsoft.AspNetCore.Http.Features.IEndpointFeature>()?.Endpoint is RouteEndpoint re)
+            {
+                operationName = re.RoutePattern.RawText;
+            }
```
stijnmoreels, line 226:
```suggestion
            if (httpContext.Features.Get<Microsoft.AspNetCore.Http.Features.IEndpointFeature>()?.Endpoint is RouteEndpoint endpoint)
            {
                operationName = endpoint.RoutePattern.RawText;
            }
```

We usually only use abbreviations in lambda's.

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
```diff
@@ -1040,21 +1040,72 @@ await using (var server = await TestApiServer.StartNewAsync(options, _logger))
...
+
+                }
+            }
+        }
+
+        [Fact]
+        public async Task GetRequest_TracksRequest_CorrectOperationNameIsLogged()
+        {
+            // Arrange
+
```
stijnmoreels, line 1059:
```suggestion
```

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
```diff
@@ -1064,21 +1115,34 @@ private static bool ContainsLoggedEventContext(InMemorySink testSink)
...
+        private static IDictionary<string, string> GetRequestLogProperties(InMemorySink testSink)
+        {
+            IEnumerable<KeyValuePair<string, LogEventPropertyValue>> properties =
+                testSink.DequeueLogEvents()
+                    .SelectMany(ev => ev.Properties);
+
+            var logEntries = properties.Where(prop => prop.Key == ContextProperties.RequestTracking.RequestLogEntry);
+            (string key, LogEventPropertyValue logEntry) = Assert.Single(logEntries);
+            var requestLogEntry = Assert.IsType<StructureValue>(logEntry);
```
stijnmoreels, line 1143: This could be reused in other methods too, right?

src/Arcus.WebApi.Tests.Integration/Logging/Controllers/EchoController.cs
```diff
@@ -20,5 +25,12 @@ public IActionResult Post([FromBody] string body)
         {
             return Ok(body);
         }
+
+        [HttpPost]
+        [Route("devices/{deviceId}/echo")]
+        public IActionResult Post([FromRoute] int deviceId, [FromBody] string body)
+        {
+            return Ok(body);
+        }
```
stijnmoreels, line 34: I'm not sure if we should pollute this simple controller with specific routes. Maybe we can create a separate API controller for this to keep this one clean and simple for other integration tests to use.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -210,7 +211,26 @@ private void LogRequest(string requestBody, string responseBody, HttpContext htt
             Dictionary<string, object> telemetryContext = CreateTelemetryContext(requestBody, responseBody, httpContext.Request.Headers, _logger);
             telemetryContext.Add("Body", "Request body is now available in 'RequestBody' dimension");
 
-            _logger.LogRequest(httpContext.Request, httpContext.Response, duration, telemetryContext);
+            var operationName = BuildOperationName(httpContext);
```
stijnmoreels, line 214: If we let the `operationName` be `null` we might consider renaming this from `Build...` to `Determine...`, `Get...` as we are not building anything anymore.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
```diff
@@ -210,7 +211,26 @@ private void LogRequest(string requestBody, string responseBody, HttpContext htt
...
+
+            _logger.LogRequest(httpContext.Request, httpContext.Response, operationName, duration, telemetryContext);
```
stijnmoreels, line 216: Should we make sure that we can still revert back to the original? (`operationName: null`) Because there may be scenario's where you do not want the route template but the exact route path?

stijnmoreels (changes requested):
A lot of small code style changes that makes the diff rather obscure. Think it was a good idea to let you implement it to get a more comfortable grip on the code base, the testing, how it works together.

This will be a valuable change, as you said for APIM perspective.
Thx!

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
fgheysels: Frankly, I personally do not see such scenario's as usable.
With API's that are very often used, you'll see an explosion of requests, simply because you log the route-parameter as part of the operation-name.

stijnmoreels: Ok,  👍

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
fgheysels: Like [this](https://github.com/arcus-azure/arcus.webapi/pull/410/commits/d788fff2e340d45fc76383811480c9b979856f88) you mean ?  
I'm not so sure about it tbh.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
fgheysels: Fixed this.

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
fgheysels: fixed

fgheysels:
> A lot of small code style changes that makes the diff rather obscure. 

Yes, a lot of changes since unnecessary whitespace has been removed automatically.

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
fgheysels: What exactly do you mean ?  It is made as a method, so it can be used in other test - methods ?

src/Arcus.WebApi.Tests.Integration/Logging/Controllers/EchoController.cs
fgheysels: I can do that, but don't know if it is necessary to introduce another controller, just to have an operation with route-params.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
stijnmoreels: No, completely `return operationName;` where `operationName: null`. Behind the scenes in Observability, this will be filled-out automatically. https://github.com/arcus-azure/arcus.observability/blob/1cef318d2afb01543d65c655e0e32462b51708e7/src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs#L163

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
stijnmoreels: Yeah, I see some duplication in the diff, so I thought that this method could be used elsewhere here as well. But could also be the code formatting that messed up the diff.

src/Arcus.WebApi.Tests.Integration/Logging/Controllers/EchoController.cs
stijnmoreels: We have several one-method controllers, all to keep the functionality and the test fixtures clear and not to pollute each other as little as possible. In this case, we could improve this since the `EchoController` is only about having a simple controller without 'complex' stuff like route parameters.

stijnmoreels:
> > A lot of small code style changes that makes the diff rather obscure.
> 
> Yes, a lot of changes since unnecessary whitespace has been removed automatically.

While I understand this is an added-value, it also means that people with other settings will have different results. I explicitly made sure that I do not do this automatic-reformat upon save, as I am sure some settings will be different on my and other people's side. And then we will have an endless PR's change of code formatting. Therefore, I think it is better to not include them in the actual PR changes.
We can leave it like this but it is a reminder for future changes.

src/Arcus.WebApi.Logging/RequestTrackingMiddleware.cs
fgheysels: Changed this.

src/Arcus.WebApi.Tests.Integration/Logging/Controllers/EchoController.cs
stijnmoreels: If the a simple controller like the `EchoController` is enough for the test, we can leave it 👍 . My concern was most of all about altering the `EchoController` for the sake of the test.

src/Arcus.WebApi.Tests.Integration/Logging/Controllers/RequestOperationNameController.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Microsoft.AspNetCore.Mvc;
+
+namespace Arcus.WebApi.Tests.Integration.Logging.Controllers
+{
+    [ApiController]
+    public class RequestOperationNameController : ControllerBase
```
stijnmoreels, line 6: Yes, this is much better 👍 Thx!

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
```diff
@@ -1040,21 +1040,71 @@ await using (var server = await TestApiServer.StartNewAsync(options, _logger))
...
+                    .Post(RequestOperationNameController.GetPostRouteWithRouteParameters(deviceId: 17))
+                    .WithJsonText("someBody");
+
+                using (HttpResponseMessage response = await server.SendAsync(request))
+                {
+                    // Assert
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                    IDictionary<string, string> requestLogProperties = GetRequestLogProperties(spySink);
+
+                    Assert.Equal("POST devices/{deviceId}/echo", Assert.Contains(OperationNameKey, requestLogProperties));
```
stijnmoreels, line 1100: This will be an improvement, indeed, to have this as operation name. Super!

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
stijnmoreels: This part is duplicated thrice now:
```csharp
 IEnumerable<KeyValuePair<string, LogEventPropertyValue>> properties =
                testSink.DequeueLogEvents()
                        .SelectMany(ev => ev.Properties);

 var eventContexts = properties.Where(prop => prop.Key == ContextProperties.RequestTracking.RequestLogEntry);
```
Which we can extract in a method.

stijnmoreels (approved):
Great stuff! 💯

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
fgheysels: I'll have a look and see if it can be improved.

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
```diff
@@ -1095,53 +1091,16 @@ await using (var server = await TestApiServer.StartNewAsync(options, _logger))
...
-            var logEntries = properties.Where(prop => prop.Key == ContextProperties.RequestTracking.RequestLogEntry);
-            (string key, LogEventPropertyValue logEntry) = Assert.Single(logEntries);
-            var requestLogEntry = Assert.IsType<StructureValue>(logEntry);
-
-            return requestLogEntry.Properties.ToDictionary(item => item.Name, item => item.Value.ToDecentString());
-        }
+        //private static bool ContainsLoggedEventContext(InMemorySink testSink)
+        //{
+        //    return testSink.HasRequestLogProperties();
+        //}
```
stijnmoreels: Great and now we can remove this!

src/Arcus.WebApi.Tests.Integration/Logging/RequestTrackingMiddlewareTests.cs
```diff
@@ -1069,9 +1065,9 @@ await using (var server = await TestApiServer.StartNewAsync(options, _logger))
                 {
                     // Assert
                     Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-                    IDictionary<string, string> requestLogProperties = GetRequestLogProperties(spySink);
+                    var requestLogProperties = spySink.GetRequestLogProperties();
```
stijnmoreels, line 1068: We should probably use the type itself here, as we describe in our code style rules, as you can't know the type beforehand here.

stijnmoreels (approved):

PR #420, opened 2022-12-19
refactor: rename the applicationId parameter to audience on the ctor of JwtTokenReader

### LINKED ISSUE #399: JwtTokenReader checks on aud instead of sub
**Describe the bug**
We use the `JwtTokenReader` class for backdoor security in our API's.  
The idea of 'backdoor security' is to verify if the incoming request on the API is originating from API Management.   

To achieve this, we add an APIM policy to create a token and send that token to the backend in an additional header.
For instance:
```xml
<authentication-managed-identity resource="target-audience" output-token-variable-name="bearerToken" ignore-error="true" />
        <set-header name="x-managed-identity-token" exists-action="override">
            <value>@((String)context.Variables["bearerToken"])</value>
        </set-header>
```
([More info](https://learn.microsoft.com/en-us/azure/api-management/api-management-authentication-policies#ManagedIdentity))

We pass the resulting token to the backend api and use the `JwtTokenReader` to check if that additional token (x-managed-identity-token) is present and valid.   By valid, we mean:  is the audience correct, as the id that we can pass in in the constructor is supposed to be an audience.

However, I believe that we should not verify the audience.  We should verify the subject of that token to perform backdoor security, as anyone can try to obtain a token for that specific audience and pass in this additional token.   We need to make sure that the token has been requested by API Management instead.

fgheysels:
It was not clear from the name of the `applicationId` parameter of the JwtTokenReader ctor how that parameter was being used.
Turns out that it was used to check if the Jwt Token matches a specific value for the audience claim.  Therefore, the parameter has been renamed to `audience`.

Closes #399

### DIFF
--- a/src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
+++ b/src/Arcus.WebApi.Security/Authorization/Jwt/JwtTokenReader.cs
@@ -31,30 +31,30 @@ namespace Arcus.WebApi.Security.Authorization.Jwt
         /// Initializes a new instance of the <see cref="JwtTokenReader"/> class.
         /// </summary>
         /// <remarks>Uses Microsoft OpenId connect discovery endpoint</remarks>
-        /// <param name="applicationId">The Azure AD Application used as audience to validate against.</param>
-        /// <exception cref="ArgumentException">Thrown when the <paramref name="applicationId"/> is blank.</exception>
-        public JwtTokenReader(string applicationId) : this(applicationId, NullLogger<JwtTokenReader>.Instance)
+        /// <param name="audience">The id of the intended audience for which this token must be validated against.</param>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="audience"/> is blank.</exception>
+        public JwtTokenReader(string audience) : this(audience, NullLogger<JwtTokenReader>.Instance)
         {
-            Guard.NotNullOrWhitespace(applicationId, nameof(applicationId), "Requires an Azure AD application ID used as audience to validate against");
+            Guard.NotNullOrWhitespace(audience, nameof(audience), "Requires an audience to validate against");
         }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="JwtTokenReader"/> class.
         /// </summary>
         /// <remarks>Uses Microsoft OpenId connect discovery endpoint</remarks>
-        /// <param name="applicationId">The Azure AD Application used as audience to validate against.</param>
+        /// <param name="audience">The id of the intended audience for which this token must be validated against.</param>
         /// <param name="logger">The logger instance to write diagnostic messages when verifying the JWT token.</param>
-        /// <exception cref="ArgumentException">Thrown when the <paramref name="applicationId"/> is blank.</exception>
-        public JwtTokenReader(string applicationId, ILogger<JwtTokenReader> logger) : this(new Dictionary<string, string> {{ JwtClaimTypes.Audience, applicationId}}, logger)
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="audience"/> is blank.</exception>
+        public JwtTokenReader(string audience, ILogger<JwtTokenReader> logger) : this(new Dictionary<string, string> {{ JwtClaimTypes.Audience, audience}}, logger)
         {
-            Guard.NotNullOrWhitespace(applicationId, nameof(applicationId), "Requires an Azure AD application ID used as audience to validate against");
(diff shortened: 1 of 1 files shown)


stijnmoreels (changes requested):
Hmm, this is technically a breaking change, if someone uses the parameter name it will break in the next version.

fgheysels:
> Hmm, this is technically a breaking change, if someone uses the parameter name it will break in the next version. Are we okay with that?

In my opinion, yes.  I agree that it is technically a breaking change, but if we leave it as is, it is actually obscure and incorrect.
Next to that, if someone specifies the parameter name, the compiler will give an error so the author is informed and it's an easy fix.   It would be different if the behavior of the implementation changes.

stijnmoreels:
Ok, 👍

stijnmoreels (approved):

gverstraete:
> > Hmm, this is technically a breaking change, if someone uses the parameter name it will break in the next version. Are we okay with that?
> 
> In my opinion, yes. I agree that it is technically a breaking change, but if we leave it as is, it is actually obscure and incorrect. Next to that, if someone specifies the parameter name, the compiler will give an error so the author is informed and it's an easy fix. It would be different if the behavior of the implementation changes.

please put this with great exclamation marks in the release notes!

Are you going to do a major version bump bc of that?

fgheysels:
> > > Hmm, this is technically a breaking change, if someone uses the parameter name it will break in the next version. Are we okay with that?
> > 
> > 
> > In my opinion, yes. I agree that it is technically a breaking change, but if we leave it as is, it is actually obscure and incorrect. Next to that, if someone specifies the parameter name, the compiler will give an error so the author is informed and it's an easy fix. It would be different if the behavior of the implementation changes.
> 
> please put this with great exclamation marks in the release notes!
> 
> Are you going to do a major version bump bc of that?

To be clear, there is no functional change here.   The name of the parameter changes but the way it is used or treated, remains the same.   The name is changed to make it more clear on what this parameter is used for.
If you specify this parameter name in your code, as in `new JwtTokenReader(applicationId: "abcd-audience")`, you'll be given a compiler error which will indicate that there is no argument called `applicationId` on the constructor.  All you have to do , is change the code to this `new JwtTokenReader(audience: "abcd-audience")`.   That's it.  No functional changes or whatsoever.

gverstraete:
> > > > Hmm, this is technically a breaking change, if someone uses the parameter name it will break in the next version. Are we okay with that?
> > > 
> > > 
> > > In my opinion, yes. I agree that it is technically a breaking change, but if we leave it as is, it is actually obscure and incorrect. Next to that, if someone specifies the parameter name, the compiler will give an error so the author is informed and it's an easy fix. It would be different if the behavior of the implementation changes.
> > 
> > 
> > please put this with great exclamation marks in the release notes!
> > Are you going to do a major version bump bc of that?
> 
> To be clear, there is no functional change here. The name of the parameter changes but the way it is used or treated, remains the same. The name is changed to make it more clear on what this parameter is used for. If you specify this parameter name in your code, as in `new JwtTokenReader(applicationId: "abcd-audience")`, you'll be given a compiler error which will indicate that there is no argument called `applicationId` on the constructor. All you have to do , is change the code to this `new JwtTokenReader(audience: "abcd-audience")`. That's it. No functional changes or whatsoever.

So potentially for your code a breaking change :) Still, same question applies :).

fgheysels:
> > > > > Hmm, this is technically a breaking change, if someone uses the parameter name it will break in the next version. Are we okay with that?
> > > > 
> > > > 
> > > > In my opinion, yes. I agree that it is technically a breaking change, but if we leave it as is, it is actually obscure and incorrect. Next to that, if someone specifies the parameter name, the compiler will give an error so the author is informed and it's an easy fix. It would be different if the behavior of the implementation changes.
> > > 
> > > 
> > > please put this with great exclamation marks in the release notes!
> > > Are you going to do a major version bump bc of that?
> > 
> > 
> > To be clear, there is no functional change here. The name of the parameter changes but the way it is used or treated, remains the same. The name is changed to make it more clear on what this parameter is used for. If you specify this parameter name in your code, as in `new JwtTokenReader(applicationId: "abcd-audience")`, you'll be given a compiler error which will indicate that there is no argument called `applicationId` on the constructor. All you have to do , is change the code to this `new JwtTokenReader(audience: "abcd-audience")`. That's it. No functional changes or whatsoever.
> 
> So potentially for your code a breaking change :) Still, same question applies :).

It's potentially a compiler error indeed and technically, we would need to bump the major version.

PR #432, opened 2023-01-20
chore: add code coverage

stijnmoreels:
Add code coverage reporting.

Relates to https://github.com/arcus-azure/arcus/issues/264


stijnmoreels:
Added `[ExcludeFromCodeCoverage]` to components that are tested via integration tests (and are also best suited to test as an integration test), for the missing parts are new unit tests added.

stijnmoreels:
> You could make sure that coverage is also reported for the integration tests. (But those tests are not ran by the CI pipeline i presume)

We do run the integration tests in the CI pipeline for every Arcus repo, but running multiple test projects with does not work (as far as I know) to combine in a single code coverage result.

fgheysels:
> > You could make sure that coverage is also reported for the integration tests. (But those tests are not ran by the CI pipeline i presume)
> 
> We do run the integration tests in the CI pipeline for every Arcus repo, but running multiple test projects with does not work (as far as I know) to combine in a single code coverage result.

I 've done this before in an Azure DevOps pipeline.  Do do that, I had to combine the reports generated by the separate test-runs and use a separate task to combine & publish them.

stijnmoreels:
> > > You could make sure that coverage is also reported for the integration tests. (But those tests are not ran by the CI pipeline i presume)
> > 
> > 
> > We do run the integration tests in the CI pipeline for every Arcus repo, but running multiple test projects with does not work (as far as I know) to combine in a single code coverage result.
> 
> I 've done this before in an Azure DevOps pipeline. Do do that, I had to combine the reports generated by the separate test-runs and use a separate task to combine & publish them.

Aha, okay, maybe that will be good here as the integration tests are actually all 'in-memory'/self-contained ones and do not require external resources.

stijnmoreels:
@fgheysels , did you have any idea how to add a threshold while combining 2 test reports? A threshold on any of the test runs does not take into account the combined results. That's a problem here as they are together above 80%, but not separate.

stijnmoreels:
This is kind of a real missing feature in Coverlet, to have a total threshold guard for merged code coverage reports.
@fgheysels , any experience on this? Together, they are more than 90% coverage, but not separately.

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,36 @@
...
+        language: [ 'csharp' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Coverlet coverage unit test
+      run: dotnet test -c Debug -p:coverletOutput=coverage.json -p:CollectCoverage=true -p:CoverletOutputFormat=\"opencover,json\" -p:ExcludeByAttribute=\"Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute\" -p:Exclude=\"[Arcus.WebApi.Tests.*]*,[Arcus.WebApi.Security]*,[Arcus.WebApi.OpenApi.Extensions]*\" src/Arcus.WebApi.Tests.Unit/Arcus.WebApi.Tests.Unit.csproj
+  
+    - name: Coverlet coverage integration test
```
stijnmoreels, line 30: Currently running the integration tests here, too. But in a later phase, we can move our CI pipeline completely to GitHub Actions so these tests will not run twice.

fgheysels (approved):

PR #442, opened 2023-03-08
chore: use appinsights pkg for isolated az func

stijnmoreels:
Use preview Application Insights NuGet package for isolated Azure Functions applications to fix the automatic dependency tracking and Arcus telemetry.

Relates to https://github.com/arcus-azure/arcus.observability/issues/359


stijnmoreels:
> Just wondering if we should look at this further: [Azure/azure-functions-dotnet-worker#1263 (comment)](https://github.com/Azure/azure-functions-dotnet-worker/issues/1263#issuecomment-1469572496)

Yes, actually hope that they answer me on this so we can use the preferred approach.

stijnmoreels:
@fgheysels , still waiting for an answer from Microsoft on how to do custom telemetry in isolated Azure Functions. Seems that this is still not really looked into.
It is the only thing stopping us from releasing the next major, here.

stijnmoreels:
A new preview is released last month: https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.ApplicationInsights/1.0.0-preview4

Seems that we do could maybe work with this package.

fgheysels:
Any update on this ?

stijnmoreels:
> Any update on this ?

Package is still on preview v0.4, two months ago. Waiting.
People at Microsoft were also saying that there is currently still a drawback when tracking telemetry. So, it works with the preview package.

stijnmoreels:
> > > Any update on this ?
> > 
> > 
> > Package is still on preview v0.4, two months ago. Waiting. People at Microsoft were also saying that there is currently still a drawback when tracking telemetry. So, it works with the preview package.
> 
> What drawback are they talking about ? 

The operation ID (our transaction ID) is not automatically filled out. Which is actually one of the whole reasons why chose to merge the Microsoft / Arcus telemetry.
I personally think that they either do not currently care enough to fix this for isolated Azure Functions, or they are planning some bigger update which telemetry will play be a bigger instead of this preview package.

> Maybe an idea to use the preview package and create a pre-release of our packge as well ? Might that be an idea ? @stijnmoreels @pim-simons ?

Until we do not know what their long-term strategy is, I honestly don't know exactly what would be best to do here.

pim-simons:
> > > Any update on this ?
> > 
> > 
> > Package is still on preview v0.4, two months ago. Waiting. People at Microsoft were also saying that there is currently still a drawback when tracking telemetry. So, it works with the preview package.
> 
> What drawback are they talking about ?
> 
> Maybe an idea to use the preview package and create a pre-release of our packge as well ? Might that be an idea ? @stijnmoreels @pim-simons ?

If their package is in preview than I think it makes sense to have a preview package on our side as well 👍🏻

stijnmoreels:
Running (again) with my head against the (tedious) Docker wall. 🤕🧱
Will have to check later why this fails exactly.

fgheysels (commented):
Don't know if we still need to do something here ?

stijnmoreels:
> Don't know if we still need to do something here ?

Also don't know 😄

stijnmoreels:
Think this is done, as it only updates the packages and related YAML files.

fgheysels (approved):

PR #460, opened 2024-02-19
chore: use new major versions of arcus packages

stijnmoreels:
Use the new major release versions of the Arcus packages Observability and Security, both with the .NET 8 support.

Relates to https://github.com/arcus-azure/arcus/issues/311


stijnmoreels:
We will first probably need to update the Messaging > Testing > Web API.

fgheysels (approved):

PR #504, opened 2026-02-05
chore(deps): use newest arcus.observability version

stijnmoreels:
Use most recent version of the Arcus.Observability package as a last hotfix release to help with migration and moving away from Arcus.WebAPI.


fgheysels (approved):
Thanks for this!  This should make life a bit easier :)

